PR #9, opened 2026-05-21
Remove the bidi Connect

### LINKED ISSUE #4: Revisit the AgentService protocol
AgentService should stop support for the bidi streaming. We no longer want sub agents to be able to trigger other agents through the bidi connection. Proto should look like:

```proto
message AgentRequest {
  string conversation_id = 1;
  string exec_id = 2;
  AgentStart start = 3;
}

message AgentResponse {
  string conversation_id = 1;
  string exec_id = 2;
  oneof type {
    AgentOutputs outputs = 3;
    AgentEnd end = 4;
  }
}

service AgentService {
  rpc Connect(AgentRequest) returns (stream AgentResponse);
}
```

The protocol will be revisited in the long term but we can keep Connect as a placeholder in its new form until we have alignment on agent resumption protocols.

rakyll:
We don't allow sub agents to use the controller to make agent handoffs anymore.

Split AgentMessage into AgentRequest and AgentResponse proto definitions.

Fixes #4.

### DIFF
--- a/examples/docker_agent/main.go
+++ b/examples/docker_agent/main.go
@@ -18,7 +18,6 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"io"
 	"log"
 	"net"
 	"os"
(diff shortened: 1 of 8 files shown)


rakyll:
cc @joycel-github @anj-s

wjjclaud (approved):

Issue #3, opened 2026-05-21
Revisit the Substrate deployment


rakyll:
The new deployment model requires:
* AX server become a MT front door
* We introduce a new binary to run the harness or the environment as an isolated actor
* We remove axepp and make Substrate calls directly from the AX server
* Allow bringing a custom container as an environment per harness on Substrate

rakyll:
Partially blocked by https://github.com/google/ax/issues/2.

rakyll:
I'm going to work on some scaffolding changes for the new deployment model, and we can split the work from there:

1. MT AX server (the new frontdoor as a ReplicaSet)
3. Placeholder HarnessService
4. axharness binary that implements the HarnessService
5. Controller2 + new logic to create/resume and suspend by conversation_id
6. Potential changes to the event log

rakyll:
See https://github.com/orgs/agentexecutor/projects/3 for tracking.

rakyll:
I'm closing this in favor of tracking via the project board.

PR #17, opened 2026-05-22
Initial commit of the Substrate refactor

rakyll:
- Added `internal/controller2/`: as a copy of the existing controller.
- Added a new deployment manifest which will replace the previous one, `internal/manifests/ax-deployment2.yaml`.
- Added `HarnessService` to proto as a placeholder.
- Added `internal/SUBSTRATE_REFACTOR_CLEANUPS.md` for some cleanups after this refactor is done.
- Added `SubstrateHarness` that can create/resume axharness as an actor with conversation ID.
- Remove unnecessary ate build tags because Substrate is now public.

Updates #3.


rakyll:
@anj-s, see internal/controller2 for the refactoring. We need it for the Substrate refractor too.

internal/controller2/controller.go
```diff
@@ -0,0 +1,218 @@
...
+	"github.com/google/ax/internal/agent"
+	"github.com/google/ax/internal/controller/executor"
+	"github.com/google/ax/internal/gemini"
+	"github.com/google/ax/internal/harness/harnesstest"
+	"github.com/google/ax/proto"
+	"github.com/google/uuid"
+)
+
+const (
+	plannerAgentID = "__planner"
```
anj-s: what do these two ids correspond to?

internal/controller2/controller.go
```diff
@@ -0,0 +1,218 @@
...
+// Controller is the main controller that coordinates all components.
+// It acts as a single-writer system for managing agentic loops.
+type Controller struct {
+	registry       *Registry
+	eventLog       executor.EventLog
+	plannerBuilder PlannerBuilder
+}
+
+// PlannerBuilder is a function that creates a PlanFunc given a Registry.
+type PlannerBuilder func(ctx context.Context, r *Registry) (agent.Agent, error)
```
anj-s: Can we call this AgentBuilder?

internal/controller2/controller.go
```diff
@@ -0,0 +1,218 @@
...
+	// TODO(jbd): Enable bringing a remote harness that implements HarnessService.
+
+	h := harnesstest.New()
+	exec, err := h.Start(ctx, req.ConversationId)
+	if err != nil {
+		return fmt.Errorf("failed to start harness session: %w", err)
+	}
+	defer exec.Close(ctx)
+
+	if err := exec.Queue(ctx, req.Inputs...); err != nil {
```
anj-s, line 80: Should queue be something that the controller owns Vs the execution interface which will be implemented by the user? I can see Run being the function that we want harness owners to implement but the queue, close, pause and resume are something will want to override from AX's side.

internal/controller2/controller.go
```diff
@@ -0,0 +1,218 @@
...
+		return fmt.Errorf("failed to queue inputs: %w", err)
+	}
+
+	hhandler := &harnessHandler{
```
anj-s, line 84: Who constructs the handler?

internal/controller2/controller.go
rakyll: Removing these, we don't need a planner anymore.

internal/controller2/controller.go
rakyll: Removed the PlannerBuilder, I think we will need something like a HarnessBuilder to initiate, it will be a follow up change as we are landing the Harness support.

internal/controller2/controller.go
rakyll: This is just an intermediate adapter to adjust the new Harness interface to the OutputHandler. We will do a big cleanup once @paramgo have the new async Exec RPC.

internal/controller2/controller.go
rakyll: It sounds like a follow up conversation for the Harness interface. We need to implement one concrete harness to figure out the right interface in my opinon.

internal/controller2/controller.go
rakyll: Added note about removing this adapter to the cleanups md.

rakyll:
PTAL

anj-s (approved):

wjjclaud (approved):

PR #22, opened 2026-05-22
feat: Integrate Antigravity harness with fallback in Controller V2

anj-s:
This change implements the integration of the Antigravity agent as a harness in AX Controller V2 as a first step to validate the surface. We are using a subprocess with input/output message as a first step.

Key changes:
1. Modified the example Antigravity agent in examples/antigravity_agent to accept prompts via command-line arguments, enabling dynamic interaction.
2. Implemented AntigravityHarness in Go (internal/harness/antigravity.go) which runs the Python agent as a subprocess, captures stdout, and streams the response. Added a TODO to transition this to a gRPC server to avoid subprocess overhead in the future.
3. Created BuildHarness in internal/controller2/builder.go to construct the appropriate harness. It performs environment checks (python3 availability and script existence) and falls back to the test harness (harnesstest) if the environment is not ready.
4. Updated Controller V2 (internal/controller2/controller.go) to use BuildHarness based on the requested AgentId, and made the Antigravity script path configurable.
5. Added a unit test in internal/controller2/controller_test.go to verify the fallback behavior when a non-existent script path is configured.

### DIFF
--- a/examples/antigravity_agent/agent.py
+++ b/examples/antigravity_agent/agent.py
@@ -1,0 +1,28 @@
+# Copyright 2026 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import asyncio
+import sys
+from google.antigravity import Agent, LocalAgentConfig
+
+async def main():
+    # Initialize the agent configuration. It automatically picks up GEMINI_API_KEY from the environment.
+    config = LocalAgentConfig()
+    async with Agent(config) as agent:
+        prompt = sys.argv[1] if len(sys.argv) > 1 else "Explain quantum computing in one sentence."
+        response = await agent.chat(prompt)
+        print(await response.text())
+
+if __name__ == "__main__":
+    asyncio.run(main())
--- a/examples/antigravity_agent/agent.py
+++ b/examples/antigravity_agent/agent.py
@@ -20,7 +20,11 @@ async def main():
     # Initialize the agent configuration. It automatically picks up GEMINI_API_KEY from the environment.
     config = LocalAgentConfig()
     async with Agent(config) as agent:
-        prompt = sys.argv[1] if len(sys.argv) > 1 else "Explain quantum computing in one sentence."
+        # Check if the user has sent a prompt otherwise thrown an error
+        
+        prompt = sys.argv[1] if len(sys.argv) > 1 else None
+        if not prompt:
+            raise ValueError("Please provide a prompt for your agent. Usage: python agent.py <prompt>")
         response = await agent.chat(prompt)
         print(await response.text())
 
--- a/internal/controller2/controller.go
+++ b/internal/controller2/controller.go
@@ -19,6 +19,7 @@ package controller2
 import (
 	"context"
 	"fmt"
+	"log"
 
 	"github.com/google/ax/internal/controller/executor"
 	"github.com/google/ax/internal/harness/harnesstest"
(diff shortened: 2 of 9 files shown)


internal/harness/antigravity.go
```diff
@@ -0,0 +1,148 @@
...
+	if e.closed {
+		return fmt.Errorf("execution is closed")
+	}
+	e.queued = append(e.queued, msg...)
+	return nil
+}
+
+// Run implements Execution.Run.
+// It executes the Python agent as a subprocess, passing the last user message as an argument.
+func (e *antigravityExecution) Run(ctx context.Context, handler Handler) error {
```
zbl94, line 90: For discussion: I am wondering what is the desired call sequence for Execution.Queue(...) and Execution.Run(...)?

If each call of Queue(...) is followed by exactly one call of Run(...), shall we merge them into one?

If multiple calls of Run(...) are needed to consume pending messages in a queue, how the Execution should notify the caller if more calls to run is needed or not?

internal/controller2/controller.go
```diff
@@ -70,7 +72,13 @@ func (d *Controller) Exec(ctx context.Context, req *proto.ExecRequest, handler E
 	// TODO(jbd): Resume an incomplete execution if there exists one.
 	// TODO(jbd): Enable bringing a remote harness that implements HarnessService.
 
-	h := harnesstest.New()
+	// Retrieve harness from registry
+	h, err := d.registry.GetHarness(req.AgentId)
```
zbl94, line 76: Looks like we are retrieving an existing harness instance given an agent id. Are we mixing the two concepts here?

internal/controller2/registry.go
```diff
@@ -237,3 +240,20 @@ func (r *Registry) Close() error {
 
 	return firstErr
 }
+// RegisterHarness registers a harness.
+func (r *Registry) RegisterHarness(id string, h harness.Harness) {
```
zbl94, line 244: As mentioned in the other thread, we can call it harnessId to be more explicit.

internal/controller2/registry.go
```diff
@@ -32,13 +33,15 @@ type Registry struct {
 	mu        sync.RWMutex
 	agents    map[string]agent.Agent
 	agentInfo map[string]*agent.AgentInfo
+	harnesses map[string]harness.Harness
```
zbl94, line 36: We are using the registry for both agents and harnesses. (ok to be not in this pr) - do  we plan to let the registry in the controller package to only manage harnesses? I think the agents managed by a harness is a harness implementation detail.

internal/harness/antigravity.go
```diff
@@ -0,0 +1,148 @@
...
+	e.mu.Lock()
+	inputs := e.queued
+	e.queued = nil
```
zbl94, line 93: The conversation history is lost here. For a poc implementation I think it's OK. But we should clarify if the controller is responsible to queue all history for the execution, or if the execution class itself is responsible to maintain the history.

internal/controller2/controller.go
anj-s: We will be consolidating both concepts since a harness and agent are the same thing. This is a temporary state where we use the same registry

internal/harness/antigravity.go
anj-s: Thats a good point to raise. I have reasons for why they should be separate. Will mark this for discussion in our meeting this week,

internal/harness/antigravity.go
anj-s: +1 This is a POC implementation where we ignore message saving on purpose.

We should not be saving any conversation history but instead maintain execution history. Will be addressing this in a follow up PR.

internal/harness/antigravity.go
zbl94: SG, added to the meeting note today.

internal/controller2/controller.go
rakyll: Agreed with some kind of consolidation. Let's add a TODO here maybe to make it clear that this is temporary.

internal/controller2/registry.go
rakyll, line 244: ditto here, i think we need some kind of cleanup.

internal/harness/antigravity.go
```diff
@@ -0,0 +1,148 @@
...
+	}
+	e.queued = append(e.queued, msg...)
+	return nil
+}
+
+// Run implements Execution.Run.
+// It executes the Python agent as a subprocess, passing the last user message as an argument.
+func (e *antigravityExecution) Run(ctx context.Context, handler Handler) error {
+	e.mu.Lock()
+	inputs := e.queued
```
rakyll, line 92: Don't we implement some kind of streaming input behavior for the queue inputs?

examples/antigravity_agent/agent.py
```diff
@@ -0,0 +1,28 @@
...
+
+import asyncio
+import sys
+from google.antigravity import Agent, LocalAgentConfig
+
+async def main():
+    # Initialize the agent configuration. It automatically picks up GEMINI_API_KEY from the environment.
+    config = LocalAgentConfig()
+    async with Agent(config) as agent:
+        prompt = sys.argv[1] if len(sys.argv) > 1 else "Explain quantum computing in one sentence."
```
rakyll: Do we need the else branch here?

Maybe it's better to handle the case prompt is "" in the harness implementation.

rakyll (commented):
I think this looks good for the first phase but leaving some comments.

internal/controller2/controller.go
```diff
@@ -69,8 +71,14 @@ func (d *Controller) Exec(ctx context.Context, req *proto.ExecRequest, handler E
 
 	// TODO(jbd): Resume an incomplete execution if there exists one.
 	// TODO(jbd): Enable bringing a remote harness that implements HarnessService.
-
-	h := harnesstest.New()
+  // TODO(anj): We need to consolidate agents and harness registration.
```
anj-s, line 74: fixing the formatting here.

internal/harness/antigravity.go
anj-s: Thats in the next PR. I wanted to validate the harness interface with a simple example.

rakyll (approved):

PR #46, opened 2026-06-02
feat: implement interactive gRPC harness client

paramgo:
Update the implementation of harness to have a go_away method.

Also, add a harnessclient to test the gRPC functionality.

### DIFF
--- a/cmd/ax/harness.go
+++ b/cmd/ax/harness.go
@@ -82,17 +82,30 @@ func NewHarnessServiceServer() *HarnessServiceServer {
 }
 
 // Connect implements the bidirectional gRPC streaming capability.
-// It receives client inputs and responds only with "Hello world".
+// It receives client inputs and responds with "hello world" unless the input message text is "go_away".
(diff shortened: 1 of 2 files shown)


cmd/ax/harness.go
```diff
@@ -82,27 +82,41 @@ func NewHarnessServiceServer() *HarnessServiceServer {
...
 			return nil
 		}
 		if err != nil {
 			return err
 		}
 
+		shouldGoAway := false
+		for _, m := range req.Messages {
+			if textBlock, ok := m.Content.Type.(*proto.Content_Text); ok {
+				if textBlock.Text.Text == "go_away" {
```
rakyll, line 101: Maybe let's add a TODO saying "go_away" will be replaced with protocol changes.

rakyll (approved):

cmd/ax/harness.go
paramgo: Done.

PR #48, opened 2026-06-04
Add the scaffolding for harness config in controller v2.

wjjclaud:
## Summary
This PR adds the scaffolding config for harness to complete controller2 path E2E. Clean up the legacy agent/planner from the controller2.
- Add a dedicated `internal/config2` package for the `-tags=harness`
  build, kept separate from the default `config` package (which retains planner/
  agent config).
- `cliutil` exports the per build tag alias: `config` in the default build;`config2` in the
  `-tags=harness` build.
- Slim `controller2` to `config2` only: drop the unused agent registry
  (`RegisterLocal/Remote/Colab/ATE`, agent maps), delete `registry_ate.go`.
- `internal/ax2.yaml` for a local config example, and `ax-deployment2.yaml` for substrate example.

## Tested
- Local: Antigravity harness works E2E:
  - `python python/antigravity/harness_server.py`
  - `go run -tags=harness ./cmd/ax exec --config internal/ax2.yaml --input "What is the weather in New York?"`
- Substrate: Now works E2E with ax-harness example. It runs on substrate with actors created, resumed and suspended. **Note**: Need [PR#49](https://github.com/google/ax/pull/49) to run E2E on substrate. Follow the README in PR#49 to test E2E.


internal/config/harnessconfig/harness_config.go
```diff
@@ -0,0 +1,124 @@
...
+}
+
+// EventLogConfig configures the event log storage.
+type EventLogConfig struct {
+	SQLiteConfig SQLiteConfig `yaml:"sqlite"`
+}
+
+// ATEConfig configures the substrate control plane endpoint used by
+// substrate harnesses.
+type ATEConfig struct {
```
rakyll: I was envisioning this differently. User shouldn't know we have a different implementation on Agent Substrate for the built-in harnesses.

We check `AX_SUBSTRATE` environment variable, if it's "1", we behave differently when serving  built-in harnesses like "antigravity". 

We'd hardcode the built-in names.

```
If AX_SUBSTRATE=1 and  harness is "antigravity", "claudecode", ...:
    - resume the harness actor for conversation_id
    - use HarnessService to communicate
    - suspend on completion
```

We'd keep ATEConfig for custom harnesses.

internal/config/harnessconfig/harness_config.go
```diff
@@ -0,0 +1,124 @@
...
+
+	"gopkg.in/yaml.v3"
+)
+
+// Config represents the main configuration for the AX harness server.
+type Config struct {
+	Server    ServerConfig    `yaml:"server"`
+	EventLog  EventLogConfig  `yaml:"eventlog"`
+	Harnesses HarnessesConfig `yaml:"harnesses,omitempty"`
+	ATE       ATEConfig       `yaml:"ate,omitempty"`
```
rakyll: Should we rename this?

``` Substrate SubstrateConfig       `yaml:"substrate,omitempty"```

File: internal/config/harnessconfig/harness_config.go [hunk 3]
```diff
@@ -0,0 +1,124 @@
...
+}
+
+// ServerConfig configures the gRPC server.
+type ServerConfig struct {
```
rakyll: ServerConfig isn't a harness config, right?

internal/config/harnessconfig/harness_config.go
```diff
@@ -0,0 +1,124 @@
...
+	Address string `yaml:"address"` // Server address to listen on (e.g., ":8494")
+}
+
+// SQLiteConfig configures the SQLite event log file.
+type SQLiteConfig struct {
```
rakyll: Ditto. SQLiteConfig or EventLogConfig aren't harness configs.

internal/config/harnessconfig/harness_config.go
```diff
@@ -0,0 +1,124 @@
...
+	// Antigravity harnesses connect to a gRPC server at a fixed address.
+	Antigravity []AntigravityHarnessConfig `yaml:"antigravity,omitempty"`
+	// Substrate harnesses are brought up as SubstrATE actors from an ActorTemplate.
+	Substrate []SubstrateHarnessConfig `yaml:"substrate,omitempty"`
+}
+
+// AntigravityHarnessConfig registers an Antigravity harness, which connects to
+// a gRPC server at a fixed address.
+type AntigravityHarnessConfig struct {
+	ID      string `yaml:"id"`      // Unique harness identifier
```
rakyll: Do we want AntigravityHarness to have a modifiable ID?

Shouldn't its ID always be "antigravity"?

internal/manifests/ax-deployment2.yaml
```diff
@@ -74,11 +74,35 @@ spec:
...
+            name: ax-server-config
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: ax-server-config
+  namespace: ax
+data:
+  ax.yaml: |
+    ate:
```
rakyll: Do we want to have an "ate" at the top level here? There are very few people on the earth that will have to override "api.ate-system.svc:443", e.g. Substrate developers. We may just use some environmental variables and consider this option when it's relevant.

internal/manifests/ax-deployment2.yaml
```diff
@@ -74,11 +74,35 @@ spec:
...
+      endpoint: "api.ate-system.svc:443"
+    harnesses:
+      default: hello-world
+      substrate:
```
rakyll, line 163: Custom harnesses shouldn't be deployed to the ax namespace. ax namespace is for ax's own deployment.

The example should be something like:

```
substrate:
        - id: hello-world
          namespace: my-harness
          template: my-harness-template
          port: 50053
```

internal/config/harnessconfig/harness_config.go
wjjclaud: Removed the ATEConfig in the update since it only contains the substrate control service's endpoint. This value is set in an env var in the manifest, and isn't supposed to be configured by users.

For now, the custom harnesses don't need it either. So removing from this PR.

internal/config/harnessconfig/harness_config.go
wjjclaud: Removed in the update.

internal/config/harnessconfig/harness_config.go
wjjclaud: My plan is to add a new config package for the new harness path, similar to `controller2`. It'll be a clean package without the legacy structs from the old config.go. 

It's a sibling package to the old config.go, and the build tag "harness" controls which package to use. So it includes the server side configs as well.

The `harnessconfig` is not a proper name for my intention. I've renamed the package to `config2` in the update. Hope it clarifies things a bit.

internal/config/harnessconfig/harness_config.go
wjjclaud: Same as above.

internal/config/harnessconfig/harness_config.go
wjjclaud: I imagine it'll be an "antigravity" type harness, with its own ID. And there could be other antigravity harnesses with different IDs and configs in the future.
E.g., "antigravity_1" with coding tools, pro model; "antigravity_2" with search tools, flash model etc.

internal/manifests/ax-deployment2.yaml
wjjclaud: Discussed offline. Moved it to an env var.

internal/manifests/ax-deployment2.yaml
wjjclaud: Updated. PTAL.

internal/ax2.yaml
```diff
@@ -0,0 +1,31 @@
...
+eventlog:
+  sqlite:
+    filename: "eventlog/log.sqlite"
+
+harnesses:
+  default: antigravity-example
+
+  antigravity:
+    - id: antigravity-example
+      address: "localhost:50053"
```
rakyll: We can set a default address:

- localhost:50053 on local deployments
- the ${ACTOR_ID}:50053 on Substrate.

rakyll (approved):
This looks good, thank you. 

We were chatting with @anj-s that let's focus on making AGY harness working and merge controller2 and config2 to replace the existing controller and config. We are getting there!

internal/ax2.yaml
wjjclaud: Yes, the address is optional. I'll update to just use the default in this example yaml.

internal/config2/config.go
```diff
@@ -0,0 +1,181 @@
...
+
+// HarnessesConfig groups harnesses to serve by type. There are two categories:
+//   - Built-in harnesses (e.g. Antigravity) whose implementation and container
+//     image are provided by AX.
+//   - Custom harnesses on substrate whose implementation and container image are
+//     provided by the user via their own ActorTemplate.
+type HarnessesConfig struct {
+	// Default is the id of the harness to serve when a request specifies no harness.
+	Default     string                     `yaml:"default,omitempty"`
+	Antigravity []AntigravityHarnessConfig `yaml:"antigravity,omitempty"`
```
anj-s, line 65: why do we need a separate harness config for Antigravity?

PR #69, opened 2026-06-08
Initial version of the harness service.

wjjclaud:
## Summary
Replaces the placeholder `HarnessMessage` with a typed, turn-based
request/response protocol for `HarnessService`, and updates all Go consumers
(substrate harness client, config2, demo stub, and test client).

This is the protocol foundation for running real harnesses (e.g. antigravity)
as substrate actors. Behavior is unchanged for end users; this is a wire-protocol
redefinition plus the Go plumbing to speak it.

Update [Issue#53](https://github.com/google/ax/issues/53)

## Changes
- **proto**: redefine `HarnessService`; regenerate Go stubs (`ax.pb.go`, `ax_grpc.pb.go`).
- **internal/harness/substrate.go**: speak the new protocol.
- **internal/config2/config.go**: pass the harness `id` into the substrate harness.
- **cmd/ax/harness.go**: rewrite the demo to the new protocol and make it
  stateful per conversation (turn counter + input recall) to show 
exercise of substrate snapshot/suspend/resume.
- **cmd/ax/harnessclient.go**: rewrite to the new protocol.

## Tested
Deploy cmd/ax/harness.go on substrate. Tested that actor behavior and response is correct.

PR #71, opened 2026-06-09
Use harness service in antigravity implementation.

wjjclaud:
## Summary
Migrates the built-in **Antigravity** harness - both the Python server and the Go
client, from the legacy `AgentService` RPC to the new `HarnessService` protocol
introduced in PR #69. Update issue #54. Also run formatter to harness_server.py.

## Tested
- Harness server: `PYTHONPATH=python:. python python/antigravity/harness_server.py` 
- e2e script `go run ./cmd/e2e`
-  ax client `go run -tags=harness ./cmd/ax exec --config internal/ax2.yaml --input "..."`


python/antigravity/harness_server.py
```diff
@@ -88,142 +99,185 @@ def hydrate_ax_history_to_steps(historical_messages) -> list[Step]:
...
+        AntigravityHarnessServiceServicer(), server
+    )
+
     listen_addr = f"{host}:{port}"
     server.add_insecure_port(listen_addr)
     print(f"Starting gRPC harness server on {listen_addr}...")
     await server.start()
     await server.wait_for_termination()
 
+
```
anj-s: There is a lot of linting fixes in this file. Can you either add the linting as a command for the team to run or revert these specific chagnes?

internal/harness/antigravity.go
```diff
@@ -107,23 +107,30 @@ func (e *antigravityExecution) Run(ctx context.Context, handler Handler) error {
...
-	// 3. Build standard AgentRequest
-	req := &proto.AgentRequest{
+	// 3. Build standard HarnessRequest.
+	start := &proto.HarnessRequest{
 		ConversationId: e.conversationID,
-		ExecId:         e.id,
-		Start: &proto.AgentStart{
-			AgentId:  "antigravity",
-			Messages: inputs,
+		HarnessId:      "antigravity",
```
anj-s, line 116: what do we use this for?

anj-s (approved):
Add TODOs for anything that is demo/testing related for us to cleanup

internal/harness/antigravity.go
wjjclaud: This is unused for now. Potentially use case is when we have multiple harness implementations in a single ax binary.

python/antigravity/harness_server.py
wjjclaud: Reverted the lint changes. Will do lint fixes in a separate PR.

PR #72, opened 2026-06-10
Add harness server readiness check via standard gRPC health.

wjjclaud:
## Summary
Closes the `ResumeActor → Connect` race in the substrate harness path by adding a
readiness gate that waits for the actor's harness to be reachable and ready
(via the standard `grpc.health.v1.Health` protocol) before connecting, and makes
both harnesses serve health.

## Tested
Both with a local antigravity server, and a substrate deployed hello-world server.

### DIFF
--- a/cmd/ax/harness.go
+++ b/cmd/ax/harness.go
@@ -30,6 +30,8 @@ import (
 	"github.com/google/ax/proto"
 	"github.com/spf13/cobra"
 	"google.golang.org/grpc"
+	"google.golang.org/grpc/health"
+	"google.golang.org/grpc/health/grpc_health_v1"
 )
 
 var (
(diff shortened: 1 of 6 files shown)


cmd/ax/harness.go
```diff
@@ -59,6 +61,11 @@ func runHarness(cmd *cobra.Command, args []string) error {
 	harnessServer := NewHarnessServiceServer()
 	proto.RegisterHarnessServiceServer(grpcServer, harnessServer)
 
+	// Serve the standard gRPC health protocol.
+	healthServer := health.NewServer()
+	healthServer.SetServingStatus("", grpc_health_v1.HealthCheckResponse_SERVING)
```
anj-s, line 66: what does grpchealth have a v1 associated with it?

anj-s (approved):

cmd/ax/harness.go
wjjclaud: v1 is the versioned API they've officially published.

Issue #55, opened 2026-06-05
Containerize + deploy antigravity on substrate


wjjclaud:
Currently, the deployment is using a dummy hello-world image https://github.com/google/ax/blob/0a79404348a47e554ac532dd0719ad1170c4b485/internal/manifests/ax-deployment2.yaml#L62

PR #80, opened 2026-06-12
Deploy Antigravity on substrate

wjjclaud:
## Summary
Replaces the hello-world Go stub for the built-in antigravity harness with the real antigravity Python harness, containerized and running as a Substrate actor. Also bumps Agent Substrate to the latest main and adapts ax to its new actor-networking model. Update #55 

## Changes
- Antigravity harness image (python/antigravity/Dockerfile, requirements.txt)
- Deploy tooling (internal/hack/install-ax.sh)
- Manifest (internal/manifests/ax-deployment2.yaml)
- Substrate dependency bump (go.mod/go.sum, cmd/axepp/main.go, internal/experimental/k8s/ate/client.go)
- ax adaptation to new actor networking (internal/config2/config.go). The worker only accepts port 80 now.
- Harness startup fix (python/antigravity/harness_server.py)

## Tested
On substrate: 
- `./internal/hack/install-ax.sh --deploy-ax-server`
- `kubectl port-forward -n ax rs/ax-server 8494:8494`
- `ax exec --input "what's the weather in NYC?"`


internal/manifests/README.md
```diff
@@ -38,9 +38,48 @@ This deploys the AX `harness` path: a built-in harness `WorkerPool` and `ActorTe
 ### 1. Build and Deploy
 
 > [!NOTE]
-> Do not manually edit `internal/manifests/ax-deployment2.yaml`. The installation script automatically injects your `${GEMINI_API_KEY}` and `${BUCKET_NAME}` environment variables during deployment.
+> Do not manually edit `internal/manifests/ax-deployment2.yaml`. The installation script automatically injects your `${GEMINI_API_KEY}`, `${BUCKET_NAME}`, and the built `${ANTIGRAVITY_IMAGE}` reference during deployment.
```
anj-s, line 41: how do we create the ANTIGRAVITY_IMAGE?

anj-s: I think I got my answer below

internal/config2/config.go
```diff
@@ -27,7 +27,7 @@ const (
 	// The substrate namespace reserved for AX's built-in harnesses.
 	defaultNamespace = "ax"
 	// The default port of HarnessService.
-	defaultPort = 50053
+	defaultPort = 80
```
anj-s: why modify this?

python/antigravity/Dockerfile
```diff
@@ -0,0 +1,61 @@
...
+# `from python.proto import ...` needs /app on the path; the generated stubs do
+# `from proto import ...`, which needs /app/python.
+ENV PYTHONPATH=/app:/app/python
+
+# Flush stdout/stderr immediately so server logs surface in container logs.
+ENV PYTHONUNBUFFERED=1
+
+EXPOSE 80
+
+CMD ["python", "-m", "python.antigravity.harness_server", \
```
anj-s, line 61: why do you need to specify this in the yaml file as well?

anj-s (approved):

internal/config2/config.go
wjjclaud: This is related to a recent change in substrate repo. They only accept port 80 as the worker port now. I think it may be better to have a separate default value for substrate workers. I'll update it.

python/antigravity/Dockerfile
wjjclaud: This is for local build and test only. Substrate ignores the CMD in Dockerfile.

anj-s: can you add a comment to that effect. thanks!

wjjclaud: Done.

PR #84, opened 2026-06-13
refactor: implement stateful Display and inject io.Writer

### LINKED ISSUE #64: Issue with model response concatenation
The responses from a single stream should be concatenated in real time as the client is updated. We are seeing it show up on separate lines

Image

anj-s:
## Summary
- Refactored `Display` in the AX CLI client to implement a state machine that tracks the type of last printed chunk (`stateNone`, `stateText`, `stateThought`). This enables real-time concatenation of streamed output blocks (text and reasoning logs) and correctly places separating newlines.
- Injected `io.Writer` interface dependency into `Display` (defaulting to `os.Stdout`).
- Eliminated global standard output redirection pipes in tests, replacing them with a local thread-safe `bytes.Buffer` wrapper.
- Enabled safe parallel execution of tests (`t.Parallel()`).
- Removed redundant and performance-blocking `os.Stdout.Sync()` calls on TTY devices.
- Cleaned up legacy unused dead code fields `loadingVisible` and `loadingStopCh`.

## Type of Change
- [x] Refactor
- [ ] Bug fix

## Related Issues
Closes #64

## Test Plan
- [x] Unit tests pass: `go test -v ./cmd/ax/internal` passes.
- [x] Integration tests pass: `go test ./...` passes.
- [x] Manually verified: Streamed model and thought chunks print concatenated on a single line sequentially, transitioning to newlines only on state changes.

### DIFF
--- a/cmd/ax/exec.go
+++ b/cmd/ax/exec.go
@@ -341,38 +341,8 @@ func runExecServer(ctx context.Context, d *internal.Display, req *proto.ExecRequ
 
 func displayContents(d *internal.Display, contents []*proto.Message) {
 	for _, output := range contents {
-		content := output.GetContent()
-		if content == nil {
-			continue
-		}
-		switch o := content.Type.(type) {
-		case *proto.Content_Text:
-			d.DisplayText(o.Text.Text)
-		case *proto.Content_Confirmation:
-			// Let the confirmation prompt handle displaying the question.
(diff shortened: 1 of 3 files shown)


rakyll:
Shouldn't we instead make `Display` take a Content argument instead of introducing stateText, stateThought, etc?

rakyll (approved):

PR #89, opened 2026-06-16
Use SubstrateHarness if AX_SUBSTRATE=1

### LINKED ISSUE #87: Introduce SubstrateHarness
Introduce a `Harness` interface implementation named `SubstrateHarness` which resumes the `ax harness` actor, communicates with it via `HarnessService`, and suspends when harness is finished working.

rakyll:
Fixes #87, #96, #95.


cmd/ax/internal/cliutil/cliutil_harness.go
anj-s, line 83: do we need this?

anj-s (approved):

cmd/ax/internal/cliutil/cliutil_harness.go
rakyll: We still need it for custom harnesses if there is any registered in the config.

Issue #120, opened 2026-06-16
Use the ax OCI image in ax-harness-workerpool


rakyll:
Today, `ax-harness-workerpool` uses the Antigravity image as the ax-harness-worker-pool.

Instead, it should do the following:

- It should use the ax image, and use `ax harness` as the startup command.
- ax image should contain the current Antigravity sidecar server and fork it
- ax image should use a more comprehensive base image (e.g. maybe Debian for now) for bash tools, skills, etc.

rakyll:
Closed via #122.

PR #121, opened 2026-06-16
Update the deployment script with ateom image and wait step.

wjjclaud:
## Summary
- Build and **pin the `ateom-gvisor` worker image** from the `go.mod` pinned substrate module, instead of letting `ko apply` compile it from the AX module — so AX's `go.mod` stays tidy. Otherwise, we would need to override the `go mod tidy` with some indirect dependencies from substrate.
- `install-ax.sh --deploy-ax-server` now **waits for the antigravity ActorTemplate to become Ready** (golden snapshot) before returning, with a small **progress spinner**.
- Bump up substrate version.

Will merge the Antigravity into ax image in a follow up for #120.

## Tested
`./internal/hack/install-ax.sh --deploy-ax-server`

### DIFF
--- a/internal/hack/install-ax.sh
+++ b/internal/hack/install-ax.sh
@@ -41,6 +41,37 @@ function log_step() {
   echo -e "${COLOR_CYAN}[step]: ${step_name}${COLOR_RESET}"
 }
 
+# wait_with_spinner runs a blocking command while showing a simple spinner on an
+# interactive terminal, then reports "done"/"failed" and returns the command's
+# exit status.
+wait_with_spinner() {
+  local msg="$1"; shift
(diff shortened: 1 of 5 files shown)


internal/hack/install-ax.sh
```diff
@@ -171,6 +202,35 @@ build_antigravity_image() {
   echo "${repo}@${digest}"
 }
 
+build_ateom_image() {
```
rakyll, line 205: Why do we need to build this image? Shouldn't we be just using whatever they publish? Only Substrate developers should be building a custom Ateom image.

rakyll (approved):
Approving this to unblock. We can revisit it later if needed.

PR #122, opened 2026-06-17
Build one comprehensive ax image for the ax-server and harness

wjjclaud:
## Summary
- Replace the separate Antigravity image with a single comprehensive **`ax` image** (full Debian + Python + the Go `ax` binary), used by **both** the ax-server (`ax serve`) and the harness actor (`ax harness`).
- `ax harness` now **forks the Antigravity Python sidecar** instead of serving the Go demo echo (which is removed).
- Removed the separate Dockerfile for antigravity.

Updates #120.

## Tested
On a cluster with substrate ready:
1. `./internal/hack/install-ax.sh --deploy-ax-server`
2. `kubectl port-forward -n ax rs/ax-server 8494:8494`
3. `ax exec --server=localhost:8494 --input="what's the weather in NYC?"`


cmd/ax/harness.go
```diff
@@ -12,30 +12,27 @@
...
-	"google.golang.org/grpc"
-	"google.golang.org/grpc/health"
-	"google.golang.org/grpc/health/grpc_health_v1"
 )
 
 var (
-	harnessPort int
+	harnessPort      int
+	harnessHost      string
+	harnessAgentFile string
```
rakyll: Name this harnessAntigravityAgentFile?

On another note, we should stop calling Antigravity an agent thought the project.

rakyll (approved):
Nice!

cmd/ax/harness.go
wjjclaud: Done.

We can have a followup for the name change in Antigravity.

cmd/ax/Dockerfile
```diff
@@ -0,0 +1,76 @@
...
+# Antigravity runtime deps, installed offline from the pre-downloaded wheels.
+COPY python/antigravity/requirements.txt /tmp/requirements.txt
+RUN --mount=type=bind,from=wheels,target=/tmp/wheels \
+    pip install --no-cache-dir --no-index --find-links=/tmp/wheels -r /tmp/requirements.txt
+
+# Python sidecar (with generated proto stubs), the agent it serves, and the ax
+# binary built in stage 1.
+COPY python/ /app/python
+COPY examples/antigravity_agent/ /app/examples/antigravity_agent
+COPY --from=build /out/ax /usr/local/bin/ax
```
rakyll: Nit: Should we copy ax into somewhere more like /ax?

wjjclaud: Done

PR #137, opened 2026-06-18
Build antigravity package directly from pip install.

wjjclaud:
- Build the `ax` image by `pip install`the Antigravity Python
  deps directly from PyPI at build time, instead of pre-downloading a linux/amd64
  wheel cache and installing it via a `wheels` build context.
- Removes the separate `--fetch-wheels` option.


rakyll (approved):

rakyll:
I'm merging this, it's blocking another issue I'm working on.

PR #139, opened 2026-06-18
Update ateomImage path in deployment manifest

rakyll:
Updating the  ateomImage to fix the failing`./hack/install-ax.sh --deploy-ax-server`.


joycel-github (approved):

rakyll:
With this change the deployment script isn't working due to many other changes. I'm merging this to be able to build the ateom image. We will soon switch to internal/hack/* version of the manifest, so we don't need to spend cycles fixing this one.

PR #156, opened 2026-06-22
feat: Support podman login for GCR/AR registries and configure runsc gvisor binaries

anj-s:
## Summary
- Added support for authenticating Podman container engine with Google Container Registry (GCR) and Artifact Registry (AR) using host-side `docker-credential-gcloud`.
- Added runsc (gVisor) binary configuration under `ActorTemplate` in GKE deployment manifest `internal/manifests/ax-deployment2.yaml` to ensure proper sandboxing.

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Breaking change

## Related Issues
*None.*

## Test Plan
- [x] Manually verified: Used to successfully build and deploy AX server/harness templates onto GKE Substrate.

## Notes for Reviewer
- Resolves authentication and runtime configuration failures when running E2E testing on remote GKE clusters using podman and gVisor.

Issue #155, opened 2026-06-22
feat: Add new content type for chunked/streaming token response in Interactions proto


anj-s:
## Context
During the implementation of gRPC token streaming buffering (Issue #146), we consolidated sequential `Text` and `Thought` tokens in the Antigravity harness server and emitted them as a single `TextContent` message. While this conforms to the Interactions proto standard and prevents invalid token-by-token `TextContent` chunks, it introduces a large user-facing latency/delay because the entire response has to be buffered before anything is sent.

## Description
To restore a fast, responsive streaming UX, we should introduce a new content type in the Interactions proto (e.g. `ChunkedText` or `TextStream`) that is designed explicitly to stream sequential token chunks.

This will allow:
1. The harness to stream token chunks instantly as they arrive.
2. The controller and client to distinguish raw streamed chunks from a final consolidated `TextContent` block.

## Proposed Tasks
- Update `proto/interactions.proto` to add a new chunked content type.
- Implement streaming chunks handling in the Go AX controller/client.
- Update the Antigravity Python SDK and harness server to emit chunks using the new type.

rakyll:
Duplicate of https://github.com/google/ax/issues/246.

PR #157, opened 2026-06-22
fix(antigravity): Buffer sequential thoughts and text tokens to prevent token-by-token streaming

### LINKED ISSUE #146: Token by token streaming shouldn't use TextContent
AGY harness streams token by token outputs by using TextContent. In Interactions proto, TextContent isn't for token streaming. Either we can introduce ContentBlock or can buffer tokens in AGY harness implementation and emit a TextContent when all tokens arrive.

anj-s:
## Summary
- Modified the Antigravity gRPC Harness server (`python/antigravity/harness_server.py`) to buffer contiguous blocks of sequential `Text` and `Thought` tokens. Consolidated tokens are yielded as a single `TextContent` message when the contiguous block ends or when a different message type (like a `ToolCall`) is received, adhering to the Interactions proto standard.
- Refactored the buffering logic in `harness_server.py` using clean local helper functions (`flush_text()`, `flush_thought()`) to eliminate code duplication across boundary switches, tool calls, and loop flushes.
- Kept GKE Substrate build and manifest changes reverted on this branch to ensure it is isolated cleanly from the infrastructure PR (#156).

## Type of Change
- [x] Bug fix
- [ ] New feature
- [x] Refactor
- [ ] Documentation
- [ ] Breaking change

## Related Issues
Closes #146

## Test Plan
- [x] Unit tests pass: Added python harness server tests (`python3 -m pytest python/antigravity/harness_server_test.py`) and ran all Go unit tests (`make test`).
- [x] Manually verified: Deployed to GKE Substrate, port-forwarded `ax-server`, and ran client queries successfully, returning consolidated responses.

## Notes for Reviewer
- Resolves token-by-token streaming bugs where individual raw tokens were incorrectly emitted in separate `TextContent` envelopes.

### DIFF
--- a/internal/hack/install-ax.sh
+++ b/internal/hack/install-ax.sh
@@ -20,11 +20,6 @@ set -o pipefail
 ROOT=$(git rev-parse --show-toplevel)
 cd "${ROOT}"
 
-# Add GOPATH/bin to PATH so that go-installed tools (like ko) can be found.
-if command -v go >/dev/null 2>&1; then
-  export PATH="${PATH}:$(go env GOPATH)/bin"
-fi
-
 if [[ -n "${PROJECT_ID:-}" ]]; then
   export KO_DOCKER_REPO="gcr.io/${PROJECT_ID}"
   echo "Using KO_DOCKER_REPO: ${KO_DOCKER_REPO}" >&2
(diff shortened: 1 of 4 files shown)


internal/hack/install-ax.sh
```diff
@@ -20,6 +20,11 @@ set -o pipefail
 ROOT=$(git rev-parse --show-toplevel)
 cd "${ROOT}"
 
+# Add GOPATH/bin to PATH so that go-installed tools (like ko) can be found.
+if command -v go >/dev/null 2>&1; then
```
rakyll: This should be in your bash profile, install-ax.sh isn't the right place for it.

This doesn't seem to the correct place.

anj-s:
Added https://github.com/google/ax/issues/155 to address the latency we will see with the buffering. This change still needs to go through to avoid using the wrong content type.

rakyll (approved):

internal/hack/install-ax.sh
anj-s: These were from another branch that should I was testing. Now the PR is scoped correctly.

PR #161, opened 2026-06-23
controller: build Gemini planner lazily when --agent is not the planner

### LINKED ISSUE #135: ax exec --agent still requires a Gemini API key
i'm trying to run a simple remote agent without a Gemini key.

i pass `--agent uppercase`, so i'd expect the planner to be skipped. but `ax serve`
still fails with "no Gemini credentials configured". so i can't run any agent
without a Gemini key, even when i pick the agent explicitly.

is this expected? i'd expect `--agent` to skip the planner.

joycel-github:
## Summary
Previously `controller.Exec` always constructed the Gemini planner before dispatching, even when the caller passed an explicit `--agent <name>`. As a result, every `ax exec --agent <name>` invocation required Gemini credentials to be configured on the local AX server, even when the user's selected agent never touched Gemini.

This change defers the Gemini planner construction until it is actually needed:
- New executions skip the planner build when `req.AgentId` is set to a concrete (non-planner) agent.
- Resumed executions build the planner only if the pending execution was originally driven by the planner (`pendingEvent.AgentId == "__planner"`).

The eager-build path is preserved via a small lazy helper, so the default behavior (`ax exec` with no `--agent`) is unchanged.

## Test coverage
Added 4 controller unit tests:
- `TestController_Exec_ExplicitAgent_SkipsPlannerBuilder` — `--agent <name>` must not call `plannerBuilder`
- `TestController_Exec_DefaultAgent_BuildsPlanner` — default behavior preserved
- `TestController_Exec_ResumePlannerExec_BuildsPlanner` — resuming planner-driven exec still builds planner
- `TestController_Exec_ResumeCustomAgentExec_SkipsPlanner` — resuming custom-agent exec must not require planner

Fixes #135

### DIFF
--- a/internal/controller/controller.go
+++ b/internal/controller/controller.go
@@ -86,6 +86,15 @@ func New(ctx context.Context, cfg Config) (*Controller, error) {
 }
 
 func (d *Controller) tryResuming(ctx context.Context, req *proto.ExecRequest, el executor.EventLog, registry map[string]agent.Agent, handler ExecHandler) (history []*proto.Message, done bool, err error) {
+	// Backwards-compatible shim: build planner eagerly if this older entry
+	// point is used (no internal callers remain after the lazy refactor).
+	return d.tryResumingLazy(ctx, req, el, registry, func() error { return nil }, handler)
+}
+
+// tryResumingLazy is like tryResuming but it asks ensurePlanner to materialize
+// the planner only if the pending execution was originally driven by the
+// planner. See Exec for context.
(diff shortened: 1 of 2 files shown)


rakyll:
You don't need to this in controller. Instead initiate the Gemini client from the given config in the Gemini agent and Gemini planner lazily.

rakyll:
We no longer need this, controller2 to controller migration fixed this issue.

PR #162, opened 2026-06-23
Add Interactions API harness

zbl94:
Adds a Harness implementation that drives an Antigravity agent through the Vertex GenAI Interactions API over HTTPS + Server-Sent Events, using the steps-based request format and the client-side ("local") environment.

The harness executes every tool the agent yields internally: built-in environment tools (view_file, run_command, list_dir, move, delete_dir, and the create/edit/multi_edit/delete_file family) run against the local filesystem and shell, while third-party function tools are dispatched to a pluggable ThirdPartyExecutor (currently a banking example until tool injection is wired). Run drives the full interaction loop to completion and streams the agent's text via the Handler.

It is exposed over the HarnessService gRPC contract via InteractionsAPIServer and the interactionsapi-harness-server command, so the harness can run as a standalone server reachable by any HarnessService client -- the substrate-ready deployment model.


joycel-github:
ready for review?

cmd/interactionsapi-harness-server/main.go
```diff
@@ -0,0 +1,77 @@
...
+	"log/slog"
+	"net"
+	"os"
+
+	"github.com/google/ax/internal/harness"
+	"github.com/google/ax/proto"
+	"google.golang.org/grpc"
+)
+
+var (
```
rakyll: Why is this an independent binary?

internal/harness/antigravityinteractions.go
```diff
@@ -0,0 +1,587 @@
...
+		}
+		eventJSON := strings.TrimSpace(strings.TrimPrefix(line, "data:"))
+		if eventJSON == "[DONE]" {
+			break
+		}
+		var event map[string]any
+		if err := json.Unmarshal([]byte(eventJSON), &event); err != nil {
+			continue
+		}
+		switch event["event_type"] {
```
rakyll, line 641: Isn't there a Go SDK for Interactions API?

internal/harness/interactionsapi_tools.go
```diff
@@ -0,0 +1,400 @@
...
+	// Declarations returns the tool declarations advertised to the agent.
+	Declarations() []functionTool
+	// Execute runs the named tool with the given arguments and returns the result
+	// value to send back to the agent (wrapped into the function_result step).
+	Execute(ctx context.Context, name string, args map[string]any) any
+}
+
+// bankingExecutor is the hardcoded placeholder ThirdPartyExecutor: a tiny
+// banking example used until the controller injects a real executor.
+type bankingExecutor struct{}
```
rakyll: Remove this.

internal/harness/interactionsapiserver.go
```diff
@@ -0,0 +1,133 @@
...
+import (
+	"context"
+	"fmt"
+
+	"github.com/google/ax/proto"
+)
+
+// InteractionsAPIServer implements proto.HarnessServiceServer backed by an
+// InteractionsAPIHarness.
+type InteractionsAPIServer struct {
```
rakyll: Remove this.

Interactions will be served from `ax harness`

rakyll (changes requested):

internal/harness/interactionsapi.go
```diff
@@ -0,0 +1,587 @@
...
+}
+
+// token obtains a bearer token from the configured TokenSource, or via gcloud.
+func (h *InteractionsAPIHarness) token(ctx context.Context) (string, error) {
+	if h.cfg.TokenSource != nil {
+		return h.cfg.TokenSource(ctx)
+	}
+	args := []string{"auth", "print-access-token"}
+	if h.cfg.ImpersonateServiceAccount != "" {
+		args = append(args, "--impersonate-service-account="+h.cfg.ImpersonateServiceAccount)
```
rakyll: Is impersonation needed for the end user impersonation? We should never rely on gcloud calls from AX.

internal/harness/antigravityinteractions.go
zbl94: Unfortunately, it seems https://pkg.go.dev/google.golang.org/genai does not support interactions yet. We can update it later to use go SDK instead.

internal/harness/interactionsapiserver.go
zbl94: Done

internal/harness/interactionsapi.go
zbl94: I removed impersonation and use `google.FindDefaultCredentials(ctx, cloudPlatformScope)` instead for default. In the harness config, the token source can be also overridden.

internal/harness/interactionsapi_tools.go
zbl94: removed.

cmd/interactionsapi-harness-server/main.go
zbl94: Per offline discussion, let's merge it into the main ax harness after harness-id based routing is done. In this pr, the separated main is removed with only the harness checked in. The testing is mainly in the e2e file which is not submitted.

internal/harness/antigravityinteractions.go
```diff
@@ -0,0 +1,638 @@
...
+
+// AntigravityInteractionsConfig configures an AntigravityInteractionsHarness.
+// Use NewAntigravityInteractionsHarness, which fills sensible defaults.
+type AntigravityInteractionsConfig struct {
+	// Endpoint is the Vertex GenAI dataplane HTTPS endpoint, e.g.
+	// "https://aiplatform.googleapis.com".
+	Endpoint string
+	// Project is the Cloud project id for the parent resource and the
+	// X-Goog-User-Project header.
+	Project string
```
rakyll: Authentication and endpoint setup should be done via the environment variables: https://github.com/google/ax#authentication

zbl94:
> ready for review?

Addressed Jaana's comments. Please take a look again

rakyll (approved):
Left a minor comment, LGTM otherwise.

internal/harness/antigravityinteractions.go
zbl94: I removed project, location, and endpoint from the harness config. Project and locations are read from env var. Endpoint should be a fix constant for customers since they should never access autopush/staging.

Issue #152, opened 2026-06-18
Set up database on k8s when AX is deployed on substrate


wjjclaud:
Set up Postgres to get started.

rakyll:
This is fixed now.

PR #163, opened 2026-06-23
feat: Support Postgres as database on substrate.

wjjclaud:
Add a durable, shared PostgreSQL event-log backend for the in-cluster ax-server, parallel to the per-pod ephemeral SQLite db. SQLite stays the default for local/dev; Postgres is used on substrate. Update #152.

## What's changed
**Event log (`internal/controller2/eventlog/`)**
- `sql.go` (new): branched from the original sqlite.go as a shared library `sqlEventLog` implementing the `EventLog` interface once over `database/sql`.
- `sql_test.go` (new): branched from the original sqlite_test.go, contains tests for shared event log logic.
- `postgres.go` (new): constructor for postgres sqlEventLog.
- `sqlite.go`: trimmed to its constructor for sqlite sqlEventLog.
- No schema changes.

**Call site**
- `EventLogBuilder` picks Postgres when `AX_SUBSTRATE` is set, otherwise SQLite.
- `harnessHandler.OnMessage` now sets `ExecResponse.Seq` from the append result (previously always `0`), so `ax exec` shows the real sequence number.

**Deployment**
- Add PostgresConfig in `config.go`.
- Add postgres configs in yaml.
- Split teardown: `--delete-ax-server` removes the server + harness but **preserves** the DB (keeps the namespace, Postgres, and PVC); `--delete-all` removes everything including the namespace/DB/PVC.

## Tested
e2e test on substrate: 
- `--deploy-ax-server` with postgres DB created.
- `ax exec` shows correct seq number.
- Query DB with `kubectl -n ax exec ax-eventlog-postgres-0 --   psql -U axuser -d axeventlog -c "SELECT conversation_id, seq FROM conversation_log ORDER BY conversation_id, seq;"`
- `--delete-ax-server`, DB is preserved and queriable.
- `--deploy-ax-server` again, and continue a previous conversation.


internal/controller2/eventlog/sql.go
```diff
@@ -0,0 +1,220 @@
...
+	defer rows.Close()
+
+	var events []*proto.ExecutionEvent
+	for rows.Next() {
+		var payload string
+		if err := rows.Scan(&payload); err != nil {
+			return nil, fmt.Errorf("eventlog: scan exec: %w", err)
+		}
+
+		ev := &proto.ExecutionEvent{}
```
joycel-github, line 134: nit: shall we throw

joycel-github (approved):

internal/controller2/eventlog/sql.go
wjjclaud: I think we can leave it as is because the execution log related code will be removed soon.

cmd/ax/internal/cliutil/cliutil_harness.go
```diff
@@ -110,6 +110,12 @@ func NewControllerFromConfig(ctx context.Context, cfg *Config) (*controller2.Con
 	return controller2.New(ctx, controller2.Config{
 		Registry: reg,
 		EventLogBuilder: func() (eventlog.EventLog, error) {
+			// Use Postgres for the event log on substrate; use SQLite for local/dev use.
+			if substrateMode {
```
rakyll: Let's not do this but have this configured from ax.yaml.

internal/controller2/eventlog/postgres.go
```diff
@@ -0,0 +1,79 @@
...
+		return nil, fmt.Errorf("postgres_eventlog: create index exec_id: %w", err)
+	}
+
+	return &sqlEventLog{db: db, lockConversation: pgAdvisoryLock}, nil
+}
+
+// pgAdvisoryLock takes a transaction-scoped advisory lock keyed on the
+// conversation ID. The lock is released automatically when the transaction
+// commits or rolls back.
+func pgAdvisoryLock(ctx context.Context, tx *sql.Tx, conversationID string) error {
```
rakyll: I'm not sure if we want this lock at the the storage driver level. Can we do a follow up on global locks outside of this PR?

internal/hack/install-ax.sh
```diff
@@ -72,7 +72,8 @@ function usage() {
   echo ""
   echo "Options:"
   echo "  --deploy-ax-server                    Build images and deploy AX server and components"
-  echo "  --delete-ax-server                    Delete AX server and components from cluster"
+  echo "  --delete-ax-server                    Delete AX server and components, preserving the event-log database"
+  echo "  --delete-all                          Delete AX server, components, and the event-log database"
```
rakyll: Let's remove --delete-all

We should never delete data from a utility script.

rakyll:
Please don't merge this change before the comments are addressed.

internal/hack/install-ax.sh
wjjclaud: Done

internal/controller2/eventlog/postgres.go
wjjclaud: Sure. Removed the lock from this PR.

cmd/ax/internal/cliutil/cliutil_harness.go
wjjclaud: Done. Added the config at the end of ax-deployment2.yaml.

rakyll (approved):

PR #176, opened 2026-06-25
Migrate controller and registry to internal/controller2

rakyll:
- Remove internal/controller
- Remove internal/config
- Remove the legacy harness build tag
- Remove the legacy ate build tag
- Use the new config

Follow ups:
- Rename controller2 to controller
- Rename config2 to config
- Remove Gemini agent and planner


wjjclaud:
Let's also remove the tag from [.github/workflows/go.yml](https://github.com/google/ax/blob/6945f06933e6ddccebdc092eb95de042c1145c48/.github/workflows/go.yml)

wjjclaud (approved):

rakyll:
Removed the tags from [.github/workflows/go.yml](https://github.com/google/ax/blob/6945f06933e6ddccebdc092eb95de042c1145c48/.github/workflows/go.yml).

PR #180, opened 2026-06-26
Add durable conversation resumability to the Antigravity Interactions harness

zbl94:
Introduce a small durable key-value store and use it to persist each conversation's interaction-chain cursor (the last interaction id), so a conversation can resume after a process restart instead of starting a new chain.

- internal/storage: a minimal Store interface (Get/Put/Delete + ErrNotFound) with documented semantics (atomicity, read-after-write, not-found-vs-error, durability) and a single-writer concurrency model. Includes a filesystem implementation (FileStore) that writes atomically via temp-file + rename.
- harness: add AntigravityInteractionsConfig.StateStore; Start loads any persisted cursor and Run persists it after each successful turn, so a fresh Execution for the same conversation continues the existing interaction chain.
- harness: document the single-writer-per-conversation expectation on the Harness interface (the controller guarantees it), which is what makes the last-write-wins store correct.

Also includes related harness improvements:
- Retry HTTP 429 (rate limit) with exponential backoff + jitter, honoring Retry-After; only 429 is retried since it is rejected before any interaction is created.
- Terminology cleanup: the within-Run FC/FR loop is the "interaction loop" (continuation turns chained via previous_interaction_id), distinct from an AX-level resume.


internal/harness/antigravityinteractions.go
```diff
@@ -115,6 +121,15 @@ type AntigravityInteractionsConfig struct {
 	// HTTPClient overrides the HTTP client. If nil, a default client with a long
 	// timeout is used.
 	HTTPClient *http.Client
+
+	// StateStore durably persists each conversation's resume cursor (the last
+	// interaction id) so a conversation can resume after a process restart or on
+	// a different replica. If nil, that cursor is kept in memory only (on the
```
joycel-github: whats the expected deployment model of AntigravityInteractionHarness? if running on substrate actor, just one worker no multiple replica story

internal/harness/antigravityinteractions.go
```diff
@@ -162,14 +177,55 @@ func NewAntigravityInteractionsHarness(cfg AntigravityInteractionsConfig) *Antig
...
-// Start implements Harness.Start.
+// resumeCursor is the small per-conversation state persisted to StateStore so a
+// conversation can resume across restarts/replicas. It records the tail of the
+// server-side interaction chain (PrevInteractionID). The cursor is only ever
+// persisted after a turn completes successfully, so a persisted cursor always
+// has a non-empty PrevInteractionID; "the conversation has started" is therefore
+// derived as PrevInteractionID != "" rather than stored separately.
+type resumeCursor struct {
+	PrevInteractionID string `json:"prev_interaction_id"`
+}
```
joycel-github: not strong opinion but what fields you expect to add under resumeCursor?

internal/storage/filestore.go
```diff
@@ -0,0 +1,127 @@
+// Copyright 2026 Google LLC
```
joycel-github: getting this state out of memory and onto disk makes sense! 

But I think we can scope this much more narrowly to the Interactions harness for now. The concrete state we need today seems to be just `conversation_id -> previous_interaction_id`. For that, I don’t think we need to introduce a general `internal/storage` abstraction yet. 

Longer term, we may want a more unified harness story story, would def love your help on it! But that probably needs input from other harnesses/use cases before we settle on the interface. 

For this PR, could we keep the persistence local to Antigravity Interactions, and name/place it as that rather than as a repo-level storage layer?

internal/harness/antigravityinteractions.go
```diff
@@ -30,29 +30,35 @@ package harness
 //     no executor is configured, no third-party tools are advertised.
```
joycel-github, line 30: with the PR, the harness allows fresh Execution continue the server-side interaction chain after restart / substrate snapshot restore.

can we add tests around this specific CUJ flow? e.g. first run starts without previous_interaction_id and persists the returned interaction id; then a new Execution for the same conversation loads it and sends it as previous_interaction_id on the next request. 

there are other CUJs but  can be taken care after milestone 1 e,g.  if we crash in the middle of a tool loop, after a function_call but before sending function_result

internal/harness/antigravityinteractions.go
zbl94: I think there can be multiple deployments of AntigravityInteractionHarness right? Because when there are many parallel conversions, a single deployment/worker can not handle. The key is for a given conversion-id, it must be also routed to the same worker (or to say, a worker resumed from the same gcs path, does not have to the same physical worker).

zbl94: I updated the comment to explore more about this deployment model.

internal/harness/antigravityinteractions.go
zbl94: I think as more advanced & powerful features in the future, we could store partial FC results (for example, 3 FC issued, 2 executed while 1 crashed, we can resume with the 2 results). Making it a struct will unlock the possibilities like this.

internal/storage/filestore.go
zbl94: As discussed in the meeting, I created a local `cursorstore` (without interface) to be only used by the Antigravity Interaction Harness

internal/harness/antigravityinteractions.go
zbl94: Added a test, antigravityinteractions_test.go

internal/harness/antigravityinteractions.go
```diff
@@ -30,25 +30,30 @@ package harness
...
+// API's previous_interaction_id; this is an implementation detail of the loop,
+// not an AX resume.
 //
-// Queue carries human input only -- the initial prompt and, in the future,
-// "steering" messages injected mid-run. It never carries tool results (the
-// harness produces those itself). Queued input is drained at each interaction
-// gap (every resume point), which is the only place the harness can influence an
-// otherwise atomic interaction.
+// Queue carries human input only -- the initial prompt and "steering" messages
+// injected mid-run. It never carries tool results (the harness produces those
```
joycel-github, line 40: shall we keep the original comment? iiuc " and "steering" messages injected mid-run." is not supported yet

internal/harness/antigravityinteractions.go
```diff
@@ -115,6 +120,14 @@ type AntigravityInteractionsConfig struct {
 	// HTTPClient overrides the HTTP client. If nil, a default client with a long
 	// timeout is used.
 	HTTPClient *http.Client
+
+	// StateDir is the directory where each conversation's resume cursor is
+	// persisted, so a conversation can resume after a restart. It is required:
+	// NewAntigravityInteractionsHarness returns an error if it is empty.
+	// Correctness relies on a single writer per conversation (the controller
+	// guarantees at most one Execution per conversation), so writes are
+	// last-write-wins.
```
joycel-github: the controller doesn't enforces this yet. can we phrase this as an expectation/requirement of the caller instead of "the controller guarantees"

internal/harness/cursorstore.go
```diff
@@ -0,0 +1,119 @@
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
```
joycel-github, line 3: nit: we can rename antigravityinteractions_cursorstore.go. Not strong on it - likely need to do a folder re-org as follow up anyway

joycel-github:
not blocking for this PR just one thing we can think of for future - today we persist the cursor after every Interactions API turn, including turns that return function calls.  it's fine to scope resumability to completed turns / next user input only. if later we want to recover crash in the middle of a tool loop, we probably need to either persist only at quiescent boundaries or store enough pending tool state.

joycel-github (approved):

internal/harness/cursorstore.go
zbl94: Yeah, I think we may need to move all things into internal/harness/antigravityinteractions/..., and other harnesses into its own folders as well to make the structure more clean. 

I will just keep cursorstore.go as it is now since it will be moved soon :-)

internal/harness/antigravityinteractions.go
zbl94: It is actually supported here! Please refer to the code at line 359-360:

```
// Fold in any human input queued (steering) since the last gap.
next = append(next, e.drainQueue()...)
```

It is not supported yet in the controller, but since the comment is for AntigravityInteractionsHarness, it should be fine.

internal/harness/antigravityinteractions.go
zbl94: Updated the comments.

PR #196, opened 2026-06-26
Remove dead code from the earlier controller

rakyll:
Remove the agent package, AgentService related proto messages, and Gemini agents in favor of the new universal agent harness. With the replacement of the old controller implementation, we no longer need agent abstractions.


joycel-github (approved):

rakyll:
Feels good to be able to merge this, please revert or follow up if it accidentally broke something.

rakyll:
Related to #183.

PR #207, opened 2026-06-27
 Minimize the AX Docker image (1.35GB -> 360MB)

rakyll:
Switch the runtime base from python:3.13 (buildpack-deps, ~1GB) to python:3.13-slim, move the Antigravity pip install into a separate builder stage so no compilers reach the final image, install only the shell tooling skills/code execution need, and drop git (~119MB).

Also replace localhost with 127.0.0.1 in case localhost is not in the hosts file.


rakyll:
See https://github.com/google/ax/pull/208 instead.

PR #208, opened 2026-06-27
Minimize the AX Docker image (1.35GB -> 360MB) for faster resumption on Substrate

rakyll:
Update Dockerfile to use python:3.13-slim and install runtime dependencies.

Revert the resolve_localhost removal. Even if /etc/hosts change is included in the Dockerfile, the entry disappears in runtime. Reverting the change to make AGY harness work on Substrate again. This is a hack and needs to be removed once we resolve the issue.

Simplify the README and the instructions for Substrate deployment.

### DIFF
--- a/cmd/ax/Dockerfile
+++ b/cmd/ax/Dockerfile
@@ -32,12 +32,17 @@ COPY . .
 RUN CGO_ENABLED=0 GOOS=${TARGETOS} GOARCH=${TARGETARCH} \
     go build -tags=harness -o /out/ax ./cmd/ax
 
-# --- Stage 2: comprehensive runtime (Debian + Python + Antigravity) -----------
-# Python image: a full Debian base (buildpack-deps) that
-# already ships bash, git, curl, wget, ca-certificates, build-essential, and the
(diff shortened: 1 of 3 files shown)


cmd/ax/Dockerfile
```diff
@@ -32,13 +32,18 @@ COPY . .
...
-# --- Stage 2: comprehensive runtime (Debian + Python + Antigravity) -----------
-# Python image: a full Debian base (buildpack-deps) that
-# already ships bash, git, curl, wget, ca-certificates, build-essential, and the
-# common CLIs (find/ps/less/grep/sed/...).
-FROM python:3.13
+FROM python:3.13-slim
 WORKDIR /ax
 
+RUN apt-get update \
+    && apt-get install -y --no-install-recommends \
```
joycel-github, line 39: do we need bash \

joycel-github (approved):

cmd/ax/Dockerfile
rakyll: Bash commands are already in the base image.

PR #210, opened 2026-06-27
Disallow registration of harnesses with empty IDs

rakyll:


rakyll:
Empty registration is used for default harness selection.

Issue #194, opened 2026-06-26
Antigravity harness read and apply configs from rpc requests


wjjclaud:
Currently, in ax/python/antigravity/harness_server.py, we maintain a map of conversation_id->agent. We need to be able to update the agent if a new config is coming from the HarnessRequest.

- Read the harness config from request.start.harness_config, and parse the bytes into JSON.
  - We need to decide what configs are accepted. E.g. model, sys instruction, mcp servers, etc.
- For new conversation, just use the config to create an agent.
- For an existing conversation_id, apply the config to the existing agent. Antigravity may not support mutation. So we probably need something like creating a new agent and inheriting the historical sessions.

rakyll:
@joycel-github  What's the status of this issue? This is one of the main blockers to release 0.2.0.

joycel-github:
done by #289 #295 #301

PR #223, opened 2026-06-29
antigravity: forward GOOGLE_CLOUD_{PROJECT,LOCATION} env to AGY config

### LINKED ISSUE #225: antigravity: PyPI 0.1.5 wheel updated 2026-06-25 with breaking API change; PR #223 broken against current PyPI
## TL;DR
PyPI `google-antigravity==0.1.5` was **re-published on 2026-06-25** with the breaking change that removed `GeminiConfig` (originally [google3 cl/930314305](http://cl/930314305), 2026-06-16). The version number did not change. `ax/python/antigravity/requirements.txt` pins `==0.1.5`, so `pip install` now retrieves the new API. **PR #223 is broken against the current PyPI wheel.**

Verified locally: `pip install --index-url https://pypi.org/simple/ google-antigravity==0.1.5` then `pytest python/antigravity/harness_server_test.py` → 9/18 tests fail with `AttributeError: 'LocalAgentConfig' object has no attribute 'gemini_config'`.

## Why we didn't see it
The dev venv (`~/ax/.venv-agy`) had the older PyPI `0.1.2` wheel installed on 2026-06-23. Force-reinstalling `0.1.5` from public PyPI reproduces the failure.

## What changed in the new wheel
- `GeminiConfig` class deleted
- `LocalAgentConfig.gemini_config` field removed
- `vertex`, `project`, `location`, `api_key` moved to top-level `LocalAgentConfig` fields (already existed as shorthand, now the only storage)
- AGY validation moved from `local_connection._validate_connection` (reading `gemini_config`) to per-endpoint `VertexEndpoint.validate_endpoint()` / `GeminiAPIEndpoint.validate_endpoint()`

## Fix for PR #223
Rewrite the two functions to use the new top-level API:

```python
def _apply_vertex_env(config) -> None:
    if not _env_use_vertex():
        return
    config.vertex = True
    config.project = os.environ.get("GOOGLE_CLOUD_PROJECT", "")
    config.location = os.environ.get("GOOGLE_CLOUD_LOCATION", "")
    # ...missing-check + print same as before, but read from config.{project,location}
```

`_has_credentials` similarly: drop the `gemini_config` indirection, read `getattr(config, "vertex", False)` / `getattr(config, "project", None)` etc. directly. Update tests accordingly.

## Open question
AGY SDK team published a breaking change under the same `0.1.5` version. Should ax pin to a wheel hash to prevent silent breakage, or coordinate with the team to bump to `0.2.0`?

joycel-github:
## Summary
The README documents `GOOGLE_CLOUD_PROJECT` / `GOOGLE_CLOUD_LOCATION` / `GOOGLE_GENAI_USE_VERTEXAI=True` as the way to run the antigravity sidecar against Vertex AI, but the AGY SDK does not read these env vars — it requires `vertex`/`project`/`location` explicitly on `LocalAgentConfig`. After [#177](https://github.com/google/ax/pull/177) bumped to `google-antigravity==0.1.5` (which restructured the config API), this surfaces as: user sets all 3 documented env vars → `ax exec` fails per-request with `A Gemini API key is required.`

This PR makes the sidecar read those env vars at startup and pass them to `LocalAgentConfig.__init__` so AGY's `@model_validator` picks them up. Also moves credential validation from per-request to startup with a single exit point in `main()` for symmetry with the vertex env validation.

## Why pass to `__init__` rather than mutate
AGY's `_apply_shorthand_configs` runs once at construction (as a `@pydantic.model_validator(mode="after")`). Post-init mutation of `config.{vertex,project,location}` does NOT regenerate `config.models`, so AGY's per-endpoint `validate_endpoint()` still sees the default `GeminiAPIEndpoint(api_key=None)` and rejects the request. Passing via `__init__(**_vertex_kwargs_from_env())` lets the validator generate the correct `VertexEndpoint(project=..., location=...)`.

## Changes
- `python/antigravity/harness_server.py`:
  - New `_env_use_vertex()` helper (env flag check).
  - New `_vertex_kwargs_from_env()` — returns `dict` of `{vertex, project, location}` from env, or `{}` if env doesn't request Vertex. Raises `ValueError` if Vertex requested but project/location missing.
  - `_build_default_config()` calls `LocalAgentConfig(..., **_vertex_kwargs_from_env())` so env values flow through AGY's `__init__` validators.
  - `_has_credentials` rewritten to mirror AGY's per-endpoint validation: only `GEMINI_API_KEY` env, `config.api_key`, or `config.vertex=True + config.{project,location}` (dropped `GOOGLE_API_KEY` and per-model api_keys; AGY doesn't accept them).
  - Credential check moved from per-request (`_run_turn`) to startup (`main`); single try/except in `main()` is the only exit point — helpers raise `ValueError`, `main` prints `ERROR: ...` to stderr and exits 1.
  - Replaced `loaded_config` module global with `_build_default_config()` factory injected into the servicer constructor. TODOs mark where #194's per-request `harness_config` layering will plug in.
  - Renamed module-private helpers to `_`-prefix (`_serve`, `_enhance_config_from_env`, `_resolve_localhost`).
- `python/antigravity/harness_server_test.py`:
  - 6 vertex env tests + 1 end-to-end `_build_default_config_picks_up_vertex_env` test + 2 credential tests (vertex requires project+location; Express Mode) + 2 servicer DI tests.
  - Existing tests updated to inject config via constructor (no module-global mutation).

## Verification

**Unit tests:** 17/17 pass against `google-antigravity==0.1.5` (current PyPI wheel, uploaded 2026-06-25). `go build` + `go test`: clean.

**End-to-end via `ax exec` → `ax serve` → sidecar → AGY → Vertex AI:**

| Scenario | Env | Result |
|---|---|---|
| Vertex happy path | `GOOGLE_GENAI_USE_VERTEXAI=True` + `GOOGLE_CLOUD_PROJECT=cloud-ai-agentic-coding` + `GOOGLE_CLOUD_LOCATION=global` | Sidecar boots: `Vertex AI backend configured: project=... location=...`. `ax exec "capital of France?"` → real Vertex/Gemini response: `Paris is the capital of France.` |
| Vertex + missing PROJECT | `GOOGLE_GENAI_USE_VERTEXAI=True` + `GOOGLE_CLOUD_LOCATION=us-east1` | Exits 1 at startup: `ERROR: Vertex AI backend requested but missing required config: project (set GOOGLE_CLOUD_PROJECT)` |
| No credentials | (none) | Exits 1 at startup: `ERROR: No Gemini credentials configured. Set GEMINI_API_KEY (AI Studio) or GOOGLE_GENAI_USE_VERTEXAI=True + GOOGLE_CLOUD_{PROJECT,LOCATION} (Vertex AI).` |

Sidecar log: `[gRPC] Running chat query` → `[gRPC] Turn completed successfully`. No mocked-tool involvement — response is real model output from Vertex AI.

## Out of scope
- Per-request `harness_config` layering (#194) — TODO markers placed.
- Substrate path: deployed via baked image / actor template, no env forwarding needed.

Closes #225.

### DIFF
--- a/python/antigravity/harness_server.py
+++ b/python/antigravity/harness_server.py
@@ -22,6 +22,7 @@ import asyncio
 import logging
 import os
 import sys
+from typing import TypedDict
 import grpc
 from grpc_health.v1 import health, health_pb2, health_pb2_grpc
 from google.protobuf.struct_pb2 import Struct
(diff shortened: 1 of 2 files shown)


python/antigravity/harness_server.py
```diff
@@ -52,59 +52,110 @@ def get_weather(city: str) -> str:
...
-    use_vertex = (
+def _env_use_vertex() -> bool:
+    """True if env requests the Vertex AI backend (vs. AI Studio API key)."""
+    return (
         os.environ.get("GOOGLE_GENAI_USE_VERTEXAI", "").lower() in ("true", "1") or
         os.environ.get("GOOGLE_GENAI_USE_ENTERPRISE", "").lower() in ("true", "1")
     )
-    if has_api_key or use_vertex:
+
+def _vertex_kwargs_from_env() -> dict:
```
rakyll: Instead of dict, would you instead introduce a type so we have some type checking.

rakyll (approved):

python/antigravity/harness_server.py
joycel-github: good point. done

PR #224, opened 2026-06-29
Change bucket var name and bump substrate version.

### LINKED ISSUE #219: Rename BUCKET_NAME to AX_SNAPSHOTS_BUCKET
In `hack/install.sh`, use `AX_SNAPSHOTS_BUCKET` environment variable. `BUCKET_NAME` is too broad.

wjjclaud:
- Rename Env var `BUCKET_NAME` → `AX_SNAPSHOTS_BUCKET` Fix #219.
- Bump substrate version and update client with new schema.

Note: Developers need to update their substrate to the pinned version after this PR.

### DIFF
--- a/internal/k8s/ate/client.go
+++ b/internal/k8s/ate/client.go
@@ -21,7 +21,9 @@ import (
 
 	"github.com/agent-substrate/substrate/pkg/proto/ateapipb"
 	"google.golang.org/grpc"
+	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/credentials/insecure"
+	"google.golang.org/grpc/status"
 )
 
 type Client struct {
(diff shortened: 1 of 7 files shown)


internal/k8s/ate/client.go
```diff
@@ -58,8 +60,11 @@ func NewClient(ns, template, target string, opts ...grpc.DialOption) (*Client, e
 // CreateActor creates a new actor.
 func (c *Client) CreateActor(ctx context.Context, id string) (*ateapipb.CreateActorResponse, error) {
 	client := ateapipb.NewControlClient(c.conn)
+	if _, err := client.CreateAtespace(ctx, &ateapipb.CreateAtespaceRequest{Name: c.namespace}); err != nil && status.Code(err) != codes.AlreadyExists {
```
rakyll, line 64: I don't think Kubernetes namespace and Atespace are the same concepts.

We need an ate space created in the deployment yaml, and use the same atespace here.

Is atespace a required field? We can keep it blank and come back to this.

rakyll (approved):
LGTM after the comment is addressed.

internal/k8s/ate/client.go
wjjclaud: Agree. This field is required by their latest commits and there's no way to configure it from yaml yet.

I'd suggest let's checkout to an earlier commit from substrate, and wait for their follow up to open the Atespace CRD in yaml.

rakyll: It's okay to merge this as it is, I'd suggest adding a TODO to revisit this.

wjjclaud: SG. Added TODO.

Issue #228, opened 2026-07-01
Remove get_weather from the Substrate instructions


rakyll:

rakyll:
Also remove the tool from the AGY harness.

PR #230, opened 2026-07-01
Remove the weather tool from the instructions

rakyll:
Weather tool was for development and testing, and is now removed.

Updates #228.


rakyll:
See https://github.com/google/ax/pull/231 instead.

PR #235, opened 2026-07-01
TUI: add interactive /config menu for per-request harness config

wjjclaud:
- **`/config` slash command** in the prompt box, with autocompletion suggestion.
- **Menu** with three options: **View/edit config**, **Load from file**, **Cancel**.
- **Per-request semantics:** the config applies to the next request only and does
  **not** carry over

Tested both locally and on substrate.


cmd/ax/harnessconfig.go
```diff
@@ -0,0 +1,141 @@
...
+}
+
+// normalizeHarnessConfigJSON trims and validates the given JSON config, returning
+// the bytes to send on the wire. Empty input clears the config (returns nil).
+func normalizeHarnessConfigJSON(s string) ([]byte, error) {
+	s = strings.TrimSpace(s)
+	if s == "" {
+		return nil, nil
+	}
+	if err := protojson.Unmarshal([]byte(s), &structpb.Struct{}); err != nil {
```
rakyll: We should turn harness_config into a bytes field. Proto encoding makes no sense outside of Google.

cmd/ax/harnessconfig.go
```diff
@@ -0,0 +1,141 @@
...
+			return nil
+		}
+	}
+}
+
+// editHarnessConfig opens the JSON editor pre-filled with the current config. It
+// returns done=true if the config was updated, or false if the user cancelled
+// back to the menu. Invalid JSON is reported and the editor re-opens with the
+// user's draft so they can fix it.
+func editHarnessConfig(d *internal.Display, harnessConfig *[]byte) (bool, error) {
```
rakyll: You have a pointer to a byte slice here. Remote it. 

Replace the argument with harnessConfig []byte

rakyll (approved):
Left comments but otherwise LGTM.

cmd/ax/harnessconfig.go
wjjclaud: Replaced protojson by json, for config format checking.

cmd/ax/harnessconfig.go
wjjclaud: Done.

PR #236, opened 2026-07-01
Make Postgres optional on substrate deployments

### LINKED ISSUE #168: Make Postgres deployment optional
Currently, `internal/hack/install-ax.sh` deploys Postgres to user's Kubernetes cluster. This is useful for testing purposes. In production use cases, users will already have a Postgres cluster and they only would like to configure the credentials (e.g. DSN) at deployment time. 

Make Posgres deployment optional in `install-ax.sh`.

wjjclaud:
### Changes
Adds a `--no-postgres` flag to `hack/install-ax.sh` so a substrate deployment can
skip the Postgres StatefulSet and run with an ephemeral **SQLite** event log
instead. Fix #168.
- manifests/ax-postgres.yaml (new): Moved out from ax-deployment.yaml. The Postgres Service + Secret +
  StatefulSet, applied only in the default (Postgres) mode.
- hack/install-ax.sh:
  - New `--no-postgres` flag.
  - Default mode: applies ax-deployment.yaml + ax-postgres.yaml with
    AX_SERVER_REPLICAS=3, resolves/generates the Postgres password, and waits
    for the Postgres StatefulSet.
  - `--no-postgres`: applies only ax-deployment.yaml with
    AX_SERVER_REPLICAS=1, rewriting the ConfigMap's eventlog block to
    sqlite: filename: "/tmp/ax-eventlog/log.sqlite".

### Tested
```
./hack/install-ax.sh --deploy-ax-server               # Postgres (default, unchanged)
./hack/install-ax.sh --deploy-ax-server --no-postgres # ephemeral SQLite
```

### DIFF
--- a/hack/install-ax.sh
+++ b/hack/install-ax.sh
@@ -73,7 +73,7 @@ function usage() {
   echo "Options:"
   echo "  --deploy-ax-server                    Build images and deploy AX server and components"
   echo "  --delete-ax-server                    Delete AX server and components, preserving the event-log database"
-  echo "  --no-postgres                         With --deploy-ax-server: use an ephemeral SQLite event log instead of Postgres"
+  echo "  --deploy-postgres                     With --deploy-ax-server: also deploy a bundled Postgres for testing (default: connect to an existing Postgres via AX_EVENTLOG_DSN)"
   echo "  -h, --help                            Show this help message"
 }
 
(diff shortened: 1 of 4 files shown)


hack/install-ax.sh
```diff
@@ -73,6 +73,7 @@ function usage() {
   echo "Options:"
   echo "  --deploy-ax-server                    Build images and deploy AX server and components"
   echo "  --delete-ax-server                    Delete AX server and components, preserving the event-log database"
+  echo "  --no-postgres                         With --deploy-ax-server: use an ephemeral SQLite event log instead of Postgres"
```
rakyll: If it's --no-postgres, it should look for credentials to connect to an existing Postgres deployment and should never rely on SQLite in Substrate deployments.

rakyll:
What about making --no-progress the default. Progress deployment is only for people who are testing this on Substrate, no real production user will use the Postgres deployment we provide from the `ax-install.sh` and want to connect to their own existing deployment.

wjjclaud:
> What about making --no-progress the default. Progress deployment is only for people who are testing this on Substrate, no real production user will use the Postgres deployment we provide from the `ax-install.sh` and want to connect to their own existing deployment.

Make sense. I updated the PR with user-provided DSN by default, and `--deploy-postgres` for testing only.

hack/install-ax.sh
wjjclaud: Done. Updated the flag name as well.

rakyll (approved):

PR #248, opened 2026-07-06
Introduce /readyz

rakyll:
To signal Substrate that the server is ready for snapshotting.

### DIFF
--- a/internal/server/server.go
+++ b/internal/server/server.go
@@ -19,12 +19,17 @@ package server
 
 import (
 	"context"
+	"errors"
 	"fmt"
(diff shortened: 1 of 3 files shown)


rakyll:
@wjjclaud is going to take over this change.

The Substrate deployments of AX is stuck in the golden snapshotting phase since the introduction of /readyz.

rakyll:
Readiness endpoint should be served from the `ax harness` server, not from the AX server.

rakyll:
Closing this, see https://github.com/google/ax/pull/254 instead.

PR #274, opened 2026-07-07
Serve the Antigravity Interactions harness over HarnessService

zbl94:
Add a Go HarnessService server for the Antigravity Interactions harness so it can run as a substrate actor, alongside the existing Python sidecar path.

- internal/harness/antigravityinteractions/server.go: Serve(...) stands up the gRPC HarnessService (plus gRPC health) and an HTTP /readyz endpoint reflecting this server's own serving state, with graceful shutdown on ctx-cancel or SIGINT/SIGTERM. Connect adapts the harness onto the wire contract: a start frame drives one turn (Start -> Queue -> Run), each agent message is streamed as a HarnessResponse{outputs} frame, and the stream terminates with exactly one HarnessResponse{end} (COMPLETED/FAILED/ CANCELED); a mid-stream cancel frame cancels the run.
- cmd/ax/harness.go: add a thin runAntigravityInteractionsHarness(ctx, systemInstructions) entrypoint that builds the config and calls Serve; the dispatch layer decides when to invoke it.
- server_test.go: bufconn tests for the start->end contract and rejection of a non-start first frame.

### DIFF
--- a/cmd/ax/harness.go
+++ b/cmd/ax/harness.go
@@ -30,6 +30,7 @@ import (
 	"syscall"
 	"time"
 
+	"github.com/google/ax/internal/harness/antigravityinteractions"
 	"github.com/spf13/cobra"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/credentials/insecure"
@@ -111,6 +112,26 @@ func runHarness(cmd *cobra.Command, args []string) error {
 		return fmt.Errorf("antigravity harness server exited: %w", err)
 	}
 	return nil
+}
+
+func runAntigravityInteractionsHarness(ctx context.Context, systemInstructions string) error {
+	if err := setHarnessWorkDir(); err != nil {
+		return err
+	}
+	cfg := antigravityinteractions.AntigravityInteractionsConfig{
+		SystemInstruction: systemInstructions,
+		StateDir:          harnessStateDir(),
+	}
+	return antigravityinteractions.Serve(ctx, cfg, harnessHost, harnessPort, harnessReadyzPort)
+}
+
+// harnessStateDir returns the directory for persisting resume cursors, defaulting
+// to the actor working directory (AX_HARNESS_WORKDIR) or the current directory.
+func harnessStateDir() string {
+	if dir := os.Getenv("AX_HARNESS_WORKDIR"); dir != "" {
+		return dir
+	}
+	return "."
 }
 
 // serveReadyz serves the HTTP /readyz endpoint on readyzPort that substrate's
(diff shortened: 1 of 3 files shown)


rakyll (approved):

internal/harness/antigravityinteractions/server.go
```diff
@@ -0,0 +1,217 @@
...
+		return fmt.Errorf("listen on %s:%d: %w", host, port, err)
+	}
+
+	grpcServer := grpc.NewServer()
+	proto.RegisterHarnessServiceServer(grpcServer, &server{h: h})
+
+	// gRPC health: report SERVING for the default service so the controller's
+	// health checks pass.
+	hs := health.NewServer()
+	hs.SetServingStatus("", healthpb.HealthCheckResponse_SERVING)
```
joycel-github, line 67: will it be okay to use hs.SetServingStatus("ax.interaction", healthpb.HealthCheckResponse_SERVING)"

zbl94: I think this is determined by the client side checking contract: in substrate.go, we have 
```
resp, err := client.Check(ctx, &grpc_health_v1.HealthCheckRequest{Service: ""})
```

wjjclaud (approved):

internal/harness/antigravityinteractions/server.go
joycel-github: thanks! not a blocker for this PR - can you help create an issue to track / change given we have differnt harness. shouldn't be a real problem given one harness at a time atm

joycel-github (approved):

cmd/ax/harness.go
```diff
@@ -116,6 +118,30 @@ func runHarness(cmd *cobra.Command, args []string) error {
...
+// in a dedicated ".ax/cursors" subdirectory UNDER the actor working directory
+// (AX_HARNESS_WORKDIR, or the current directory) so AX's internal state is kept
+// out of the agent-visible workspace while still residing within the actor's
+// snapshotted filesystem (so it survives snapshot/restore).
+func harnessStateDir() string {
+	base := os.Getenv("AX_HARNESS_WORKDIR")
+	if base == "" {
+		base = "."
+	}
+	return filepath.Join(base, ".ax", "cursors")
```
rakyll: We may consider a different path for this.

Working directory is where the harness is operating and it often reads and edits the files underneath even if they are hidden. I think somewhere under ~/.ax/ is more appropriate by default and we can make it configurable from ax.yaml.

rakyll:
You can merge this and I can work on some changes for the routing layer. `ax harness` should start a server that implements the `HarnessService` and route by `harness_id`.

cmd/ax/harness.go
zbl94: I updated it to use `~/.ax/` by default. @wjjclaud is working on the plumbing from the controller/substrate side and we plan to use `systemInstruction` as the first example field. Once we have that, adding a `stateDir` should be very easy.

Issue #270, opened 2026-07-07
Use Go embedding features to embed the python directory into the binary


rakyll:
In runtime, extract the AGY sidecar and set things up without user having to know about the underlying sidecar.

rakyll:
Fixed via https://github.com/google/ax/pull/276.

PR #251, opened 2026-07-06
Auto-start Antigravity harness sidecar from ax exec (#249)

joycel-github:
Fixes #249.

## Problem

`ax exec --input "..."` (README quickstart) fails with `connection refused` because the Antigravity Python sidecar on `127.0.0.1:50053` has no owner in local mode — users have to know to start `ax harness` first, which is Hidden and undocumented.

## Design: identify the sidecar via gRPC health service name

Rather than blindly assuming `:50053` is our sidecar (fragile — flagged by @rakyll in review), we probe the configured endpoint using the gRPC health protocol with a fully-qualified service name `harness.antigravity` (aligned with the `harnesses.antigravity` key in `ax.yaml`).

Three outcomes:

| Probe result | Interpretation | Action |
|---|---|---|
| `SERVING` for `harness.antigravity` | Existing AGY sidecar | **Reuse** — attach to it (Sidecar.Forked=false; user owns its lifetime) |
| `NOT_FOUND` / `NOT_SERVING` / `Unimplemented` | Some other gRPC service on this port | **Error** — explicit message asking user to update `harnesses.antigravity.endpoint` in ax.yaml |
| Connection refused | Port free | **Fork** Python sidecar; wait for it to become healthy; tie its lifetime to the controller |

Python sidecar registers `harness.antigravity` as a named health entry alongside the existing overall `""` entry. Go side probes the named entry.

## Config knobs

- `harnesses.antigravity.endpoint` in `ax.yaml` (default `127.0.0.1:50053`) — user's contract; port must be exclusively for the AGY sidecar. `SidecarConfig.Host`/`Port` in Go come from parsing this field.
- `AX_ANTIGRAVITY_NO_AUTOSTART=1` — disable autostart entirely (testing, or users who want to manage the sidecar themselves without any probing)

## Non-goals (out of scope, tracked separately)

- **Not shipping Python** with `go install ax` — this PR assumes the user has `python3` and `google-antigravity` installed (i.e. cloned the repo and `pip install -r python/antigravity/requirements.txt`). Zero-install quickstart is a separate CUJ.
- **`ax harness` behavior unchanged** — still the Docker/Substrate entrypoint.
- **Remote endpoints** (non-loopback) don't autostart — you can't fork a Python process on another machine.

## Follow-up issues opened

- **#263** — Migrate substrate `/readyz` health probe from `""` to `harness.antigravity`, then remove the redundant `""` entries in `harness_server.py` + `harnesstest.go`.
[...]

Issue #249, opened 2026-07-06
AGY harness fails with connection error


rakyll:
When run locally, AGY harness fails. We require user to start the AGY harness service but it shouldn't be their responsibility. We should fork the server from the AX binary.

```
ax exec
Conversation: 4bc74292-2fc9-4a24-b767-5bbcf91a9424

⏺ Hello

Error: error executing with local server: harness execution failed: failed to call gRPC HarnessService.Connect: rpc error: code = Unavailable desc = connection error: desc = "transport: Error while dialing: dial tcp 127.0.0.1:50053: connect: connection refused"
Error: error executing with local server: harness execution failed: failed to call gRPC HarnessService.Connect: rpc error: code = Unavailable desc = connection error: desc = "transport: Error while dialing: dial tcp 127.0.0.1:50053: connect: connection refused"
```

joycel-github:
Related PRs:
- #273  — introduces the `pythonsidecar` package that powers the autofork machinery.
- #275  — wires the autofork into `ax exec` via `pythonsidecar`.

rakyll:
Fixed via https://github.com/google/ax/pull/276.

PR #34, opened 2026-05-28
Add gRPC Streaming for Harness interface and integration

anj-s:
Add stateful gRPC streaming input/output and streaming AGY example.

### 1. AX V2 Controller Integration (Registry-Centric Design)
*   Introduces `internal/controller2` containing the new single-writer orchestrator.
*   Integrates `harness.Harness` and `harness.Execution` interfaces to abstract agent sandboxing and execution pipelines.
*   Implements a robust Registry-centric harness lookup with runtime fallbacks.

### 2. Antigravity gRPC Streaming Harness (`internal/harness/antigravity.go`)
*   Moves the Antigravity SDK integration away from one-off subprocess execution to an always-on, stateful gRPC protocol.
*   Go client connects via standard gRPC dials, constructs the strongly-typed `AgentRequest`, and streams output callbacks directly to V2.

### 3. Python gRPC Harness Server (`python/antigravity/harness_server.py`)
*   Created a python gRPC server implementation conforming to `pb.AgentService` and using `grpc.aio.server()`.
*   Natively unpacks messages list, hydrates conversation history, and yields `AgentResponse` frames (thoughts, text, tool calls).
*   **Architectural Documentation Added**: Included a detailed `NOTE ON ARCHITECTURE` documenting the generic loader design where `harness_server.py` dynamically imports agent files (defaulting to `agent.py`) using `--agent_file` flags to provide clean L1/L2 SDK boundary decoupling.

### 4. Python Agent Scaffolding (`examples/antigravity_agent/agent.py`)
*   Defines custom Python tools (`get_weather`), model identity, and persona configuration.
*   Provides a dual-purpose standalone CLI sandbox for L2 API local debugging.
*   **Architectural Documentation Added**: Included a detailed `NOTE ON ARCHITECTURE` explaining how this file acts as both a standalone L2 CLI debugging tool and a declarative config module imported by `harness_server.py`.

### 5. E2E Setup & Running Instructions (`cmd/e2e/main.go`)
*   Consolidated Go harness commands and established the main E2E verification runner (`cmd/e2e/main.go`).
*   **Instructions Added**: Appended explicit, step-by-step headers to `cmd/e2e/main.go` documenting how to start the Python server (with proper `PYTHONPATH` to bypass python path collisions) and drive the Go client demonstration.

### 6. Test Coverage & Verification
*   Go unit tests in `internal/harness/antigravity_test.go` (Mock gRPC server, Passed).
[...]

PR #275, opened 2026-07-08
Auto-start Antigravity Python sidecar from ax exec

joycel-github:
Progresses #249 (autofork done; PYTHONPATH & pip install still user responsibility — see Scope + Known limitation)

## Scope

This PR only wires the sidecar startup via `pythonsidecar`. It does **not** install the Python runtime dependencies of `python/antigravity/harness_server.py`. Users are expected to have them installed before running `ax exec`:

```
pip install -r python/antigravity/requirements.txt
```

## Known limitation: set `PYTHONPATH` before `ax exec`
Local users need to manually prepend the repo root and `python/` to `PYTHONPATH`:
```bash
export PYTHONPATH="$PWD:$PWD/python:$PYTHONPATH"
```
Without this, the sidecar fork fails with:
```
ImportError: cannot import name 'content_pb2' from 'proto' (unknown location)
```
**Root cause.** `harness_server.py` uses `from python.proto import ax_pb2` (needs repo root on `sys.path`) while generated `ax_pb2.py` uses `from proto import content_pb2` (needs `python/` on `sys.path`). Python's default `sys.path` satisfies only one. This has been the case since PR #34 introduced the streaming harness.
**Substrate is unaffected.** `cmd/ax/Dockerfile:64` sets `ENV PYTHONPATH=/ax-app:/ax-app/python`, so pods inherit it via the container image.
**Follow-up.** `go:embed`-ing `python/` into the ax binary and pointing `PYTHONPATH` at the extracted cache dir. Out of scope for this PR.

## Summary

`ax exec` now forks the Antigravity Python sidecar automatically via `pythonsidecar` (#273). Users no longer need to run `ax harness` in a separate terminal.

## Design rationale

This PR is intentionally minimal. An earlier attempt (#251, 1122 lines) shipped the full CUJ end-to-end — identity probe to distinguish AGY vs foreign sidecars, `Pdeathsig` for parent-crash cleanup, a shared autostart package for future harnesses, and an `AX_ANTIGRAVITY_NO_AUTOSTART` escape hatch. Post-review discussion converged on stacking the work as smaller PRs: land the minimum autofork here, add each capability as its own reviewable follow-up. This PR is that minimum. Known follow-ups (all `TODO`'d in code or tracked as issues):

- Reuse an existing AGY sidecar via a named gRPC health service identity probe (currently: any port-in-use fails the fork)
- Parent-crash cleanup (`Pdeathsig` on Linux)
- Auto PYTHONPATH via `go:embed` (#270 / addressed by #276)
- Auto `pip install` at first run

## API

- `antigravity.New(ctx, address, autoStart)` — when `autoStart=true`, forks the sidecar, waits for `TCPReady`, installs a signal handler that stops it on SIGINT/SIGTERM.
- `cliutil.NewControllerFromConfig` passes `autoStart=true` in local mode.

Substrate mode and the e2e demo pass `autoStart=false` (unchanged behavior).

## Tests
1. unit test 
2. manual run 

(.venv) lhuan@lhuan:~/ax-test$ ./ax exec --input "hi"
Starting gRPC harness server on 127.0.0.1:50053...
Conversation: 7b7974ff-0b58-445b-a836-ca41e3500f43

⏺ hi

[gRPC] Connect turn requested. conv_id=7b7974ff-0b58-445b-a836-ca41e3500f43
[gRPC] Creating new Agent instance for conv_id=7b7974ff-0b58-445b-a836-ca41e3500f43
[gRPC] Running chat query: hi
[gRPC] Turn completed successfully.
Hello! How can I help you today?
seq=2

### DIFF
--- a/cmd/ax/internal/cliutil/cliutil_test.go
+++ b/cmd/ax/internal/cliutil/cliutil_test.go
@@ -23,31 +23,6 @@ import (
 	"github.com/google/ax/internal/config"
 )
 
-func TestNewControllerFromConfig_DefaultHarness(t *testing.T) {
-	cfg := &config.Config{
-		EventLog: config.EventLogConfig{
-			SQLiteConfig: config.SQLiteConfig{
-				Filename: filepath.Join(t.TempDir(), "log.sqlite"),
-			},
-		},
-		Harnesses: config.HarnessesConfig{
-			Antigravity: config.AntigravityHarnessConfig{
-				Default:  true,
-				Endpoint: "localhost:50053",
-			},
-		},
-	}
-
-	c, err := NewControllerFromConfig(context.Background(), cfg)
-	if err != nil {
-		t.Fatalf("NewControllerFromConfig: %v", err)
-	}
-	if c == nil {
-		t.Fatal("expected non-nil controller")
-	}
-	c.Close()
-}
-
 func TestNewControllerFromConfig_BuiltinSubstrate(t *testing.T) {
 	t.Setenv("AX_SUBSTRATE", "1")
 
--- a/cmd/ax/internal/cliutil/cliutil.go
+++ b/cmd/ax/internal/cliutil/cliutil.go
@@ -56,6 +56,12 @@ func NewControllerFromConfig(ctx context.Context, cfg *Config) (*controller.Cont
 	// substrate actors ("1").
 	substrateMode := os.Getenv("AX_SUBSTRATE") == "1"
 
+	// Validate config before local-mode antigravity.New forks the Python
+	// sidecar, so a config error surfaces without spawning a subprocess.
+	if len(cfg.Harnesses.Substrate) > 0 && !substrateMode {
+		return nil, fmt.Errorf("custom substrate harnesses require AX_SUBSTRATE=1")
+	}
+
 	// Built-in harnesses.
 	var defaultHarnessID string
 	var antigravityHarness harness.Harness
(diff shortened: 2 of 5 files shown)


cmd/ax/internal/cliutil/cliutil_test.go
```diff
@@ -48,6 +58,38 @@ func TestNewControllerFromConfig_DefaultHarness(t *testing.T) {
...
+s.listen()
+while 1: s.accept()
+`
+	if err := os.WriteFile(filepath.Join(pkg, "harness_server.py"), []byte(src), 0644); err != nil {
+		t.Fatalf("write stub: %v", err)
+	}
+	t.Setenv("PYTHONPATH", root)
+}
+
+func getFreePort(t *testing.T) int {
```
rakyll: Let's remove this. By the time you want to open the connection, that port may not be free. While opening the connection, you will get an error if the port is not free.

cmd/ax/internal/cliutil/cliutil_test.go
```diff
@@ -48,6 +58,38 @@ func TestNewControllerFromConfig_DefaultHarness(t *testing.T) {
 	c.Close()
 }
 
+// stubPythonAntigravityModule writes a minimal python/antigravity/harness_server.py on PYTHONPATH that binds the requested port and blocks, so pythonsidecar.TCPReady succeeds without the real AGY SDK.
+func stubPythonAntigravityModule(t *testing.T) {
```
rakyll: See my offline comment, we should remove this.

rakyll (approved):
LGTM after the comments are addressed.

rakyll:
LGTM

PR #276, opened 2026-07-08
Embed python artifacts into ax

rakyll:
This change embeds the python/ directory into the ax binary.

- At runtime, it extracts the python/ directory to `~/.ax/python`.
- It runs, `pip install -r requirements.txt`
- And starts the AGY harness service.

This is a hack until `antigravityinteractions` package is ready to replace `antigravity`.

Tested:

- [x]  Locally
- [x]  On Substrate

### DIFF
--- a/internal/pythonsidecar/sidecar.go
+++ b/internal/pythonsidecar/sidecar.go
@@ -44,8 +44,6 @@ type Config struct {
 	// ReadyFunc is an optional function to check if the server is ready to accept requests.
 	// When provided, Start will poll ReadyFunc until it returns nil or the context expires. (Optional)
 	ReadyFunc func(ctx context.Context) error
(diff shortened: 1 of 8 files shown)


wjjclaud (approved):

internal/pythonsidecar/sidecar.go
```diff
@@ -83,15 +80,32 @@ func (s *Sidecar) Start(ctx context.Context) error {
 		return fmt.Errorf("Module cannot be empty")
 	}
 
-	// Prepare arguments: python3 -u -m module [args...]
+	// Prepare arguments: python -u -m module [args...]
```
joycel-github, line 83: nit python3?

joycel-github (approved):

PR #277, opened 2026-07-08
Antigravity interactions harness registry and config

wjjclaud:
Wires the built-in Antigravity Interactions harness into the controller for #271.
- Config (internal/config): new antigravity interactions harness config.
- Registry wiring (cmd/ax/internal/cliutil): registers for both local mode and substrate mode.
- Default state dir: move the default state dir helper to the shared package.
- Example yaml commented out for enabling in followups.

### Tested
On local mode, uncomment config in `ax.yaml`
`export GOOGLE_CLOUD_PROJECT=<project_ID>`
`go run ./cmd/ax exec --harness antigravity-interactions --input "hello"`

### Follow up
Routing in `ax harness` for substrate mode.

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -96,10 +96,9 @@ type AntigravityHarnessConfig struct {
 // AntigravityInteractionsHarnessConfig registers the built-in Antigravity
 // Interactions harness (over the Vertex GenAI Interactions API).
 type AntigravityInteractionsHarnessConfig struct {
-	Default           bool   `yaml:"default,omitempty"`            // Default harness or not
-	Agent             string `yaml:"agent"`                        // Interactions API agent
-	SystemInstruction string `yaml:"system_instruction,omitempty"` // Optional system prompt
-	StateDir          string `yaml:"state_dir,omitempty"`          // Resume-cursor directory (optional; defaults to ~/.ax/cursors)
+	Default  bool   `yaml:"default,omitempty"`   // Default harness or not
+	Agent    string `yaml:"agent,omitempty"`     // Interactions API agent (default: antigravityinteractions.DefaultAgent)
+	StateDir string `yaml:"state_dir,omitempty"` // Resume-cursor directory (optional; defaults to ~/.ax/antigravityinteractions/cursors)
 }
 
 // SubstrateHarnessConfig registers a custom harness deployed on substrate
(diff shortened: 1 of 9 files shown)


internal/config/config.go
```diff
@@ -89,6 +93,15 @@ type AntigravityHarnessConfig struct {
 	Endpoint string `yaml:"endpoint,omitempty"` // HarnessService address
 }
 
+// AntigravityInteractionsHarnessConfig registers the built-in Antigravity
+// Interactions harness (over the Vertex GenAI Interactions API).
+type AntigravityInteractionsHarnessConfig struct {
+	Default           bool   `yaml:"default,omitempty"`            // Default harness or not
+	Agent             string `yaml:"agent"`                        // Interactions API agent
+	SystemInstruction string `yaml:"system_instruction,omitempty"` // Optional system prompt
```
rakyll: We should remove SystemInstruction from here. It should be supplied per request.

internal/config/config.go
```diff
@@ -89,6 +93,15 @@ type AntigravityHarnessConfig struct {
 	Endpoint string `yaml:"endpoint,omitempty"` // HarnessService address
 }
 
+// AntigravityInteractionsHarnessConfig registers the built-in Antigravity
+// Interactions harness (over the Vertex GenAI Interactions API).
+type AntigravityInteractionsHarnessConfig struct {
+	Default           bool   `yaml:"default,omitempty"`            // Default harness or not
+	Agent             string `yaml:"agent"`                        // Interactions API agent
```
rakyll: This isn't required, right? It's only needed if the user wants to override.

internal/config/config.go
```diff
@@ -89,6 +93,15 @@ type AntigravityHarnessConfig struct {
...
+	SystemInstruction string `yaml:"system_instruction,omitempty"` // Optional system prompt
+	StateDir          string `yaml:"state_dir,omitempty"`          // Resume-cursor directory (optional; defaults to ~/.ax/cursors)
```
rakyll: Maybe we should use ~/.ax/antigravityinteractions/state to make it more clear that it's state associated with the AGY Interactions implementation.

rakyll (approved):
Left some comments but LGTM otherwise.

internal/config/config.go
wjjclaud: Done

internal/config/config.go
wjjclaud: Done. Now optional with default value.

internal/config/config.go
wjjclaud: Updated to ~/.ax/antigravityinteractions/cursors.

Issue #246, opened 2026-07-06
antigravity harness server - support streaming 


joycel-github:
## Problem statement

Antigravity harness currently has slow TTFT  for `Text` / `Thought` output.

The harness currently buffers contiguous text/thought chunks and only emits after the block completes. This makes the content boundary clear for downstream display code, but it also means AX cannot render the first token until the whole buffered block is done.

Desired behavior:

- emit the first visible output as soon as the model stream produces it
- preserve the difference between answer text and thought-summary text
- keep final/durable content semantics correct
- avoid a broad protocol redesign

## Proposal

Add explicit streaming delta content types:

- `TextDeltaContent`
- `ThoughtSummaryDeltaContent`

These are modeled after the streaming primitives `TextDelta` and `ThoughtSummaryDelta` in [`google-genai`](https://github.com/googleapis/python-genai).

## Streaming semantics

The harness **should not** infer sentence boundaries.

Instead, deltas are ordered append-only chunks for the current contiguous output run. A delta run is complete when:

- the next delta has a different content kind
- the role changes
- a non-delta output arrives, such as tool call/tool result
- the stream ends

Responsibilities:

- Harness server emits deltas immediately and preserves order.
- Controller accumulates adjacent deltas with the same role/kind for durable logs/history.
- Display renders deltas immediately and switches formatting when role/kind changes.

## Implementation tasks

1. **Proto**
   - Add `TextDeltaContent` and `ThoughtSummaryDeltaContent`, modeled after Antigravity SDK primitives.
   - Add `text_delta` and `thought_summary_delta` to `Content`.

2. **Antigravity harness**
   - Emit `text_delta` / `thought_summary_delta` as chunks arrive.
   - Stop waiting for the full contiguous text/thought block before yielding.

3. **Controller**
   - Accept delta content in stream handling.
   - Accumulate adjacent deltas into durable text/thought output where needed.
   - Flush accumulation on role change, kind change, non-delta output, or stream end.

4. **Display / clients**
   - Render `text_delta` as streamed assistant text.
   - Render `thought_summary_delta` as streamed thought-summary text.
   - Add handling for the new delta types wherever content variants are switched on.
   **- Note Display / clients side are reasonable for formatting**

rakyll:
We can move this to v0.3.0 milestone.

rakyll:
[...]

PR #281, opened 2026-07-09
proto: add TextDeltaContent and ThoughtSummaryDeltaContent (task 1 of #246)

joycel-github:
Task 1 (Proto) of #246 — antigravity harness streaming support.

## What

- Add `TextDeltaContent` and `ThoughtSummaryDeltaContent` modeled after the `TextDelta` / `ThoughtSummaryDelta` streaming primitives in [google-genai](https://github.com/googleapis/python-genai).

## Follow-up PRs will be sent in sequence 

More details see issue #246 

- **Task 2 — Antigravity harness:** emit `text_delta` / `thought_summary_delta` as chunks arrive from the antigravity sdk 
- **Task 3 — Controller:** accept delta content in stream handling; accumulate adjacent deltas into durable text/thought output where needed; flush on role change, kind change, non-delta output, or stream end. (`internal/controller/`)
- **Task 4 — Display / clients:** render `text_delta` as streamed assistant text and `thought_summary_delta` as streamed thought-summary text; add handling for the new delta types wherever content variants are switched on. Note that display is resonsible for identifying the boundary of different content and format accordingly (primary site: `cmd/ax/internal/display.go`).

### DIFF
--- a/proto/content.pb.go
+++ b/proto/content.pb.go
@@ -156,7 +156,7 @@ func (x ImageContent_MimeType) Number() protoreflect.EnumNumber {
 
 // Deprecated: Use ImageContent_MimeType.Descriptor instead.
 func (ImageContent_MimeType) EnumDescriptor() ([]byte, []int) {
-	return file_proto_content_proto_rawDescGZIP(), []int{10, 0}
+	return file_proto_content_proto_rawDescGZIP(), []int{12, 0}
 }
 
 type AudioContent_MimeType int32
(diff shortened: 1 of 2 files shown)


rakyll:
These fields are not Interactions Content compatible, am I right? We shouldn't introduce Content types that are not represented in Interactions.

joycel-github:
we have decided to go with interaction proto and this issue will be postponed to milestone 3 when the proto is stable

PR #285, opened 2026-07-09
Harden Antigravity Interactions harness HTTP client

zbl94:
Fix a substrate suspend/resume failure and tighten the config surface:

- Disable HTTP keep-alives on the harness's default client. On substrate the actor is suspended after a turn and resumed for the next (with a new routable IP), which leaves any pooled keep-alive connection stale; reusing it made the next turn's request fail. Opening a fresh connection per request avoids that at the cost of an extra handshake.
- Remove the public HTTPClient override from AntigravityInteractionsConfig. External callers don't configure the harness's transport, and exposing it risked silently reintroducing the keep-alive bug. New now always owns its (keep-alive-disabled) client; tests inject a fake transport via an unexported newWithHTTPClient seam.
- Reorder AntigravityInteractionsConfig fields to list required (Agent, StateDir) before optional, with clearer doc comments.

Build and package tests pass.

### DIFF
--- a/internal/harness/antigravityinteractions/antigravityinteractions.go
+++ b/internal/harness/antigravityinteractions/antigravityinteractions.go
@@ -211,6 +211,10 @@ func newWithHTTPClient(cfg AntigravityInteractionsConfig, hc *http.Client) (*Ant
 		// for the next (with a new routable IP), which leaves any pooled keep-alive
 		// connection stale; reusing it makes the next turn's request fail. Opening a
 		// fresh connection per request avoids that at the cost of an extra handshake.
+		//
+		// Clone DefaultTransport (rather than a bare &http.Transport{}) so we inherit
+		// its production defaults -- proxy from the environment, dial/TLS/handshake
+		// timeouts, and HTTP/2 -- and only override keep-alives.
 		tr := http.DefaultTransport.(*http.Transport).Clone()
 		tr.DisableKeepAlives = true
 		hc = &http.Client{Timeout: 10 * time.Minute, Transport: tr}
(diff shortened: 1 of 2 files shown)


internal/harness/antigravityinteractions/antigravityinteractions.go
```diff
@@ -191,15 +192,28 @@ type AntigravityInteractionsHarness struct {
...
-	hc := cfg.HTTPClient
 	if hc == nil {
-		hc = &http.Client{Timeout: 10 * time.Minute}
+		// Disable HTTP keep-alives so no TCP/TLS connection is reused across
+		// requests. On substrate the actor is suspended after a turn and resumed
+		// for the next (with a new routable IP), which leaves any pooled keep-alive
+		// connection stale; reusing it makes the next turn's request fail. Opening a
+		// fresh connection per request avoids that at the cost of an extra handshake.
+		tr := http.DefaultTransport.(*http.Transport).Clone()
+		tr.DisableKeepAlives = true
```
rakyll, line 219: We should create a new Transport here instead of cloning and mutating the old one.

```
tr := &http.Transport{}
```

rakyll (approved):
Left a comment. LGTM after it's addressed.

internal/harness/antigravityinteractions/antigravityinteractions.go
zbl94: Thanks! I kept `Clone()` here rather than `&http.Transport{}` on purpose:
`http.DefaultTransport` carries production defaults we want — `Proxy:
ProxyFromEnvironment`, dial/TLS/handshake timeouts, `ForceAttemptHTTP2`, and
idle-conn settings — and a bare `&http.Transport{}` would silently drop all of
those (notably env-proxy support and dial timeouts, which matter for the actor's
networking). So I clone it and only override `DisableKeepAlives`, and added a
comment explaining that. Happy to switch to an explicit transport that
re-specifies those defaults instead if you'd prefer.

wjjclaud:
Tested with @zbl94 on substrate. This fixed the connection issue on substrate.

wjjclaud (approved):

Issue #269, opened 2026-07-07
design harnessConfig contract


joycel-github:
## Problem
`Harness.Start(ctx, conversationID, harnessConfig []byte)` is passed
through but no harness consumes it. No shared vocabulary, no version,
no per-adapter scope rules.

## Proposal
1. **Classify every field by lifecycle layer** — `service-startup` /
   `session-sticky` / `execution-per-turn` / `runtime-protocol`.
   Layer determines adapter behavior.
2. **Keep `bytes` on the wire.** Parse to a versioned Go struct
   (`schema_version` handshake) inside the adapter. No proto churn.
3. **HITL is runtime protocol, not config** — approvals, questions,
   answers flow as events, never as fields.

## Phases
1. **Design** — finalize proto/struct shape.
2. **Implement pilot harness: Antigravity SDK** — parse + wire into
   the Antigravity Python sidecar. First real consumer; establishes
   the sticky-fingerprinting pattern.
3. Extend to `AntigravityInteractionsHarness`.
4. HITL bridge (runtime events).

rakyll:
Isn't this a duplicate of https://github.com/google/ax/issues/194? Can you please update the existing bug? I'm closing the duplicate.

Issue #203, opened 2026-06-27
Set Antigravity `save_dir` to substrate's persistent mount


joycel-github:
Antigravity SDK defaults `save_dir` to `tempfile.mkdtemp(prefix="antigravity_")` → `/tmp/antigravity_xxx/{sdk_conv_id}.db`. This works as substrate snapshot includes `/tmp`.

Better design: once substrate team lands the persistent home dir mount, set `save_dir` to that path rather than rely on implicit `/tmp` snapshotting.

Not milestone 1.

rakyll:
@joycel-github This is fixed, right?

PR #289, opened 2026-07-09
Replace Antigravity sidecar in-memory Agent cache with SDK-native resume

joycel-github:
## Problem

Antigravity harness caches Agent per conversation in memory. Substrate snapshot won't include memory, and local server restarts when a conversation resumes. This setup doesn't work for conversation
resumption.

## Assumption 

1. Single writer so no lock needed 

## Fix

1. Delete the cache. 
3. Use `LocalAgentConfig`'s native `(conversation_id,save_dir)` — SDK persists trajectory state to
`{save_dir}/{sdk_conv_id}.db`. 
4. **Note:** Each AX conversation gets its own directory under `~/.ax/antigravity/conversations/{ax_conv_id}/`; on
subsequent turns we discover the SDK's hash-id from the `.db` filename
and pass it back to resume.

**This is because SDK's `conversation_id` is resume-only — passing a non-existent id errors.** Hence discover-then-pass, not pass-always. 

Alternatively we can store the {AX convs_id : AGY convs_id} separately but this simplify the overall logic and setup 

## Next

`save_dir` is hardcoded here. Should be injected by controller via `harness_config` — see #269, #203. TODO in code.

## Tasks

- [x] Test local (`ax exec --conversation <id>` across fresh sidecar
      processes → resume works, isolation works)
- [x] Test substrate

### DIFF
--- a/python/antigravity/harness_server.py
+++ b/python/antigravity/harness_server.py
@@ -21,6 +21,7 @@ import argparse
 import asyncio
 import logging
 import os
+import pathlib
 import re
 import sys
 from typing import TypedDict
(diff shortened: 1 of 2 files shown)


python/antigravity/harness_server.py
```diff
@@ -226,110 +220,108 @@ async def _run_turn(self, request):
...
+                print(f"[gRPC] Running chat query: {latest_query_text}")
+                response = await conversation.chat(latest_query_text)
             
-            # To avoid streaming individual tokens inside TextContent messages (which is not
-            # supported by the Interactions proto/TextContent specifications), we buffer
-            # contiguous blocks of text and thought chunks, yielding them only when the 
-            # contiguous block ends or a different chunk type is received.
-            text_chunks = []
-            thought_chunks = []
+                # To avoid streaming individual tokens inside TextContent messages (which is not
```
joycel-github, line 231: starting here until line 354 are just difference caused by format. No real change

wjjclaud: Maybe we can have a base PR to just format the python? I feel it'll be more clear.

rakyll (approved):

python/antigravity/harness_server.py
joycel-github: the differences were caused because 1) i deleted quite some codes so they got shifted up 2) I added  "async with Agent(per_conv_config) as agent" per request so some codes got additional Indentation 

no other format change

wjjclaud (approved):

joycel-github:
tested substrate with full snapshot. lgtm

PR #292, opened 2026-07-10
Antigravity sidecar state dir from config

joycel-github:
Follow-up to #289.
## Problem
Sidecar's trajectory storage directory was hardcoded in Python. No way
for users to relocate it.
## Fix
Plumb `state_dir` through: `ax.yaml` → `AntigravityHarnessConfig` → Go
client → sidecar `--state-dir` arg. Mirrors the existing
`antigravity-interactions.state_dir` pattern.
Python default (`~/.ax/antigravity/conversations`) kept as transitional
fallback for substrate mode (ActorTemplate doesn't inject the flag
yet); removable once it does.
## Tasks
- [x] Test local
- [ ] Test substrate

### DIFF
--- a/cmd/ax/internal/cliutil/cliutil.go
+++ b/cmd/ax/internal/cliutil/cliutil.go
@@ -71,8 +71,9 @@ func NewControllerFromConfig(ctx context.Context, cfg *Config) (*controller.Cont
 		if address == "" {
 			address = "127.0.0.1:50053"
 		}
-		// Local mode: the harness owns the Python sidecar.
-		antigravityHarness, err = antigravity.New(ctx, address, true)
+		// Local mode: the harness owns the Python sidecar. Empty StateDir
+		// means the sidecar applies its own default
+		antigravityHarness, err = antigravity.New(ctx, address, cfg.Harnesses.Antigravity.StateDir, true)
 		if err != nil {
 			return nil, fmt.Errorf("antigravity harness: %w", err)
 		}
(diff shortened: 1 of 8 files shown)


cmd/ax/internal/cliutil/cliutil.go
```diff
@@ -71,8 +71,9 @@ func NewControllerFromConfig(ctx context.Context, cfg *Config) (*controller.Cont
 		if address == "" {
 			address = "127.0.0.1:50053"
 		}
-		// Local mode: the harness owns the Python sidecar.
-		antigravityHarness, err = antigravity.New(ctx, address, true)
+		// Local mode: the harness owns the Python sidecar. Empty StateDir
+		// means the sidecar applies its own default
+		antigravityHarness, err = antigravity.New(ctx, address, cfg.Harnesses.Antigravity.StateDir, true)
```
rakyll, line 76: Users shouldn't know about state_dir. It's an AGY SDK implementation detail that shouldn't be transparent to AX users.

Instead use the `config.AXAssetsDir()` to generate a state_dir.

rakyll (changes requested):

rakyll:
I'm approving this to unblock the other changes, but let's address my comment in a follow up.

rakyll (approved):

cmd/ax/internal/cliutil/cliutil.go
joycel-github: sure - will create a follow up to remove it from ax.yaml for both this and interaction harness

PR #293, opened 2026-07-10
Use GOOGLE_CLOUD_PROJECT variable and remove PROJECT_ID usage

### LINKED ISSUE #290: Use GOOGLE_CLOUD_PROJECT instead of PROJECT_ID in install-ax.sh
`GOOGLE_CLOUD_PROJECT` is the conventional environment variable for GCP projects, leverage the convention rather than introducing a new environment variable.

rakyll:
Fixes https://github.com/google/ax/issues/290.


wjjclaud:
We also need to update the reference in `ax/manifests/ax-deployment.yaml` and `ax/internal/harness/antigravityinteractions/antigravityinteractions.go`.

wjjclaud (approved):

rakyll:
Done!

PR #295, opened 2026-07-11
parse harness config from request

joycel-github:
Parse `HarnessStart.harness_config` (delivered to the sidecar as json in bytes )
into a JSON object and overlay it onto the server's `LocalAgentConfig` before each turn.

Minimum parsing and validation logic in the adapter as intended  
- Only blocks a request from overriding AX-owned setup (`save_dir`,
  `conversation_id`); these are injected last so a request can't redirect
  trajectory storage.
- No distinction between request-time vs conversation-creation overrides
  -- intentionally left to the harness behind the adapter.
- No validation yet for fields only meaningful to a runtime request, not
  harness_config (e.g. confirmation). Rejected today since
  `LocalAgentConfig` doesn't accept them.

Config is reconstructed (not `model_copy`'d) so the SDK re-validates and
surfaces its own error; invalid config fails the turn with
`INVALID_ARGUMENT` instead of crashing.

## Next
- Error on `LocalAgentConfig` unknown fields (pydantic silently drops
  them today; TODO in code).
- Merge the `state_dir` change from #292.

## Verification
Local `ax exec --harness-config-json`: valid override applied, AX-owned
field rejected, SDK validation error surfaced, malformed JSON rejected,
resume intact across a per-turn override.

### DIFF
--- a/python/antigravity/harness_server.py
+++ b/python/antigravity/harness_server.py
@@ -19,6 +19,7 @@
 import argparse
 
 import asyncio
+import json
 import logging
 import os
 import pathlib
@@ -39,6 +40,16 @@ from python.proto import content_pb2
 from google.antigravity import Agent, AgentConfig, LocalAgentConfig
 from google.antigravity.types import Text, Thought, ToolCall
 
+
+# AX owns these fields for resumption; a request must not set them.
+# TODO: add validation for fields that are unsafe to set per execution
+# (e.g. credentials, deployment routing) or that may only be set at
+# conversation creation.
+_AX_MANAGED_CONFIG_FIELDS = frozenset({"conversation_id", "save_dir"})
+
+
+class HarnessConfigError(ValueError):
+    """Raised when request harness_config is not a valid overlay."""
 
 
 class VertexKwargs(TypedDict, total=False):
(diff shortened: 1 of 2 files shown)


python/antigravity/harness_server.py
```diff
@@ -40,6 +41,17 @@
 from google.antigravity.types import Text, Thought, ToolCall
 
 
+# AX owns these fields for resumption; a request must not set them.
+# TODO: add validation for fields that are unsafe to set per execution
+# (e.g. credentials, deployment routing) or that may only be set at
+# conversation creation.
+_AX_MANAGED_CONFIG_FIELDS = frozenset({"conversation_id", "save_dir"})
```
rakyll, line 48: Why do we need conversation_id? It's already in the request.

python/antigravity/harness_server.py
```diff
@@ -136,6 +148,30 @@ def _has_credentials(config: AgentConfig | None) -> bool:
 
     return False
 
+def _parse_harness_config(raw_config: bytes) -> dict[str, object]:
+    if not raw_config:
+        return {}
+
+    try:
+        config = json.loads(raw_config.decode("utf-8"))
```
rakyll: Can we introduce a type for the config? It's very hard to read/manage it if it's represented as `dict[str, object]`.

python/antigravity/harness_server.py
```diff
@@ -150,17 +186,29 @@ class AntigravityHarnessServiceServicer(ax_pb2_grpc.HarnessServiceServicer):
...
+    def _build_config_for(
+        self, conversation_id: str, harness_config: bytes = b""
+    ) -> LocalAgentConfig:
+        # Request overlay first, then AX-managed persistence values last so a
+        # request can never redirect trajectory storage.
+        overrides = _parse_harness_config(harness_config)
+
+        # TODO(#269, #203): should be injected by the controller.
+        overrides["save_dir"] = str(
+            pathlib.Path.home() / ".ax" / "antigravity" / "conversations" / conversation_id
```
rakyll: The path here won't work on Windows. Use the platform specific path separator.

rakyll (approved):
LGTM after the comments.

python/antigravity/harness_server.py
joycel-github: this is to prevent end user incorrectly set conversation_id or save_dir in request.harness_config.

wjjclaud (approved):

python/antigravity/harness_server.py
joycel-github: harnessConfig is JSON encoded in bytes so decoding it produces a dict. I updated the impl so the dict is used as an intermediate oly inside _build_config_for; method boundary and returned contract are the SDK config LocalAgentConfig hence. This avoids duplicating the SDK schema in a seperate TypedDict

python/antigravity/harness_server.py
joycel-github: pathlib.Path.home() and / operator are platform-aware. Rebased on the state_dir change from #292 so path is  derived from the injected self._state_dir.

PR #300, opened 2026-07-13
Add GEAP Skill Registry integration for local harnesses

zbl94:
Materialize agentskills.io skills from the GEAP Skill Registry (Vertex AI v1beta1) into an on-disk skills directory before a harness starts, so the Antigravity and Antigravity Interactions harnesses can use registry-hosted skills on the local `ax exec` / `ax serve` path.

- internal/skills/materialize: harness-agnostic package that reads config.SkillsConfig, drives the registry client (ListSkills / GetSkill / GetSkillRevision / skills:retrieve), safe-unzips payloads to <target_dir>/<skill-id>/, and reports what it wrote. First-wins on duplicate ids with a warning; fail-safe (a registry error never blocks harness startup).
- config: harnesses.<id>.skills.registries[] with per-registry selection (skills / query / all), required target_dir, and a validated "exactly one selection mode" rule. Also wires the interactions harness's system_instruction from ax.yaml.
- cliutil: materializes skills at controller construction; for the interactions harness (no SKILLS_DIR concept) it appends a discovery pointer to the system instruction.

Scope: local path only; the substrate/pod path does not yet read ax.yaml. Verified end-to-end against a live registry (by-id and by-query).

### DIFF
--- a/cmd/ax/internal/cliutil/cliutil.go
+++ b/cmd/ax/internal/cliutil/cliutil.go
@@ -26,6 +26,7 @@ import (
 	"github.com/google/ax/internal/harness/antigravity"
 	"github.com/google/ax/internal/harness/antigravityinteractions"
 	"github.com/google/ax/internal/harness/substrate"
+	"github.com/google/ax/internal/skills/geminienterprise"
 )
 
 // Controller is the active controller type for this build.
@@ -64,6 +65,22 @@ func NewControllerFromConfig(ctx context.Context, cfg *Config) (*controller.Cont
 	var defaultHarnessID string
 	var err error
 
+	// Materialize registry skills once, up front (skills config is top-level and
+	// harness-agnostic; each actor runs a single harness that consumes the
+	// materialized folder). Unconditional when configured. Fail-safe: a registry
+	// error degrades capability but never blocks harness creation. The
+	// interactions harness (no SKILLS_DIR concept) is told where the materialized
+	// skills are via a pointer appended to its system instruction. Only the local
+	// path materializes; substrate/pod does not yet read ax.yaml.
+	//
+	// TODO(joycel): wire the Antigravity SDK harness too. It discovers skills via
+	// SKILLS_DIR, so its SKILLS_DIR needs to be pointed at the materialized
+	// target_dir; currently only the interactions harness is fully wired.
+	var skillsPointer string
+	if !substrateMode {
+		skillsPointer = antigravityinteractions.SkillsSystemInstruction(geminienterprise.Materialize(ctx, cfg.Skills))
+	}
+
 	// Built-in Antigravity harness.
 	var antigravityHarness harness.Harness
 	if !substrateMode {
@@ -98,7 +115,8 @@ func NewControllerFromConfig(ctx context.Context, cfg *Config) (*controller.Cont
 	// Built-in Antigravity Interactions harness.
 	var antigravityInteractionsHarness harness.Harness
 	if !substrateMode {
-		agent := cfg.Harnesses.AntigravityInteractions.Agent
+		aiCfg := cfg.Harnesses.AntigravityInteractions
+		agent := aiCfg.Agent
 		if agent == "" {
 			agent = antigravityinteractions.DefaultAgent
 		}
@@ -108,9 +126,12 @@ func NewControllerFromConfig(ctx context.Context, cfg *Config) (*controller.Cont
 		if sErr != nil {
 			return nil, fmt.Errorf("antigravity-interactions harness: %w", sErr)
 		}
+		// skillsPointer was built once, up front, from the top-level skills
+		// config (see above). Append it to any configured system instruction.
 		antigravityInteractionsHarness, err = antigravityinteractions.New(antigravityinteractions.AntigravityInteractionsConfig{
-			Agent:    agent,
-			StateDir: stateDir,
+			Agent:             agent,
+			SystemInstruction: antigravityinteractions.JoinSystemInstruction(aiCfg.SystemInstruction, skillsPointer),
+			StateDir:          stateDir,
 		})
 	} else {
 		antigravityInteractionsHarness, err = substrate.New(config.AntigravityInteractionsHarnessID, "", "", config.AntigravityInteractionsTemplate, 80)
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -48,6 +48,11 @@ type Config struct {
 	Server    ServerConfig    `yaml:"server"`
 	EventLog  EventLogConfig  `yaml:"eventlog"`
 	Harnesses HarnessesConfig `yaml:"harnesses,omitempty"`
+	// Skills sources skills from the Gemini Enterprise Skill Registry into on-disk folders
+	// before the harness starts. It is harness-agnostic: each actor runs exactly
+	// one harness, which consumes the materialized folder(s). Optional; disabled
+	// when no registry is enabled.
+	Skills    SkillsConfig    `yaml:"skills,omitempty"`
 	Telemetry TelemetryConfig `yaml:"telemetry,omitempty"`
 }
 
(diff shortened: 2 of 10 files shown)


wjjclaud (commented):
For the package name, is it better to use something like `skills/geap` or `skills/geminienterprise` to call out it is specific to skills from GEAP?

zbl94:
> For the package name, is it better to use something like `skills/geap` or `skills/geminienterprise` to call out it is specific to skills from GEAP?

I moved these to internal/skills/gemini/client.go, and called the package gemini since go prefers short package names

joycel-github (approved):
LGTM with a few comments here and there. Thanks vm for getting this in!

cmd/ax/internal/cliutil/cliutil.go
```diff
@@ -64,6 +65,20 @@ func NewControllerFromConfig(ctx context.Context, cfg *Config) (*controller.Cont
...
+	// materialized folder). Unconditional when configured. Fail-safe: a registry
+	// error degrades capability but never blocks harness creation. The result
+	// feeds every local harness built below: the Antigravity SDK harness
+	// auto-loads its SKILLS_DIR (no pointer needed), while the interactions
+	// harness (no SKILLS_DIR concept) is told where skills are via a pointer
+	// appended to its system instruction. Only the local path materializes;
+	// substrate/pod does not yet read ax.yaml.
+	var skillsPointer string
+	if !substrateMode {
+		skillsPointer = skillsSystemInstruction(geminienterprise.Materialize(ctx, cfg.Skills))
```
joycel-github: mind leaving a TODO here for Antigravity SDK? it uses SKILLS_DIR only atm. Happy to address from my side later

cmd/ax/internal/cliutil/skills.go
```diff
@@ -0,0 +1,59 @@
...
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/google/ax/internal/skills/geminienterprise"
+)
+
+// skillsSystemInstruction builds a system-instruction pointer telling the agent
+// where its materialized skills live and lists them. This is harness-specific
```
joycel-github: understood that this is put under cliutil to allow other harnesses to use it in the future as needed 

with a second thought - would be good if we can move this skillsSystemInstruction and joinSystemInstruction to internal/harness/antigravityinteractions given 

1. only interaction harness needs it. 
2. system instruction / discovery logic should be harness specific. 

in the future when we have use case that harness share same discovery / instruction, we can have dedicated common class that support that which prob should be outside of cliutil

cmd/ax/internal/cliutil/cliutil.go
zbl94: Done

cmd/ax/internal/cliutil/skills.go
joycel-github: mostly wanna make sure cli doesn't cover harness-specific prompt / discovery semantics

internal/skills/geminienterprise/unzip.go
```diff
@@ -0,0 +1,195 @@
...
+	if remaining < 0 {
+		return 0, fmt.Errorf("unzipped size cap exceeded before %q", f.Name)
+	}
+	rc, err := f.Open()
+	if err != nil {
+		return 0, fmt.Errorf("opening entry %q: %w", f.Name, err)
+	}
+	defer rc.Close()
+
+	out, err := os.OpenFile(target, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, 0o644)
```
joycel-github: this is the right permission?

cmd/ax/internal/cliutil/skills.go
zbl94: moved the skills helper under interaction harness

internal/skills/geminienterprise/unzip.go
joycel-github: /script will have executable script that we will need to preserve the permission

internal/skills/geminienterprise/client.go
```diff
@@ -0,0 +1,400 @@
...
+	payloadB64, revision, err := c.fetchPayload(ctx, ref)
+	if err != nil {
+		return nil, err
+	}
+	zipped, err := base64.StdEncoding.DecodeString(payloadB64)
+	if err != nil {
+		return nil, fmt.Errorf("decoding payload: %w", err)
+	}
+	skillDir := filepath.Join(targetDir, ref.SkillID)
+	// Replace any prior materialization of this skill so stale files don't linger.
```
joycel-github, line 231: this remove old version first and then unzip / materialize new version iiuc - if remove succeeds and subsequent fails, what would happen

internal/skills/geminienterprise/client.go
```diff
@@ -0,0 +1,400 @@
...
+	}
+	if err := os.MkdirAll(targetDir, 0o755); err != nil {
+		return nil, fmt.Errorf("registry: creating target dir %q: %w", targetDir, err)
+	}
+
+	res := &fetchResult{}
+	for _, ref := range refs {
+		// First-wins: skip (with a warning) any id already materialized into this
+		// dir, rather than overwriting it.
+		if won, byRegistry := claimed.claim(targetDir, ref.SkillID, regIdx); !won {
```
joycel-github, line 180: should claim happen at the end

internal/skills/geminienterprise/materialize.go
```diff
@@ -0,0 +1,169 @@
...
+		refs := make([]skillRef, 0, len(rc.Skills))
+		for _, s := range rc.Skills {
+			refs = append(refs, skillRef{SkillID: s.ID, Revision: s.Revision})
+		}
+		return selection{SkillRefs: refs}
+	}
+	if rc.Query != nil && strings.TrimSpace(rc.Query.Text) != "" {
+		return selection{Query: rc.Query.Text, TopK: rc.Query.TopK}
+	}
+	return selection{All: true}
```
joycel-github, line 163: imo, all should only be used when it's explicitly set given the potential download of overloading. also let's do a trimSpace to avoid " " accidentally fall into All

internal/skills/geminienterprise/client.go
```diff
@@ -0,0 +1,400 @@
...
+	refs, err := c.resolveSelection(ctx, sel)
+	if err != nil {
+		return nil, fmt.Errorf("registry: resolving selection: %w", err)
+	}
+	if err := os.MkdirAll(targetDir, 0o755); err != nil {
+		return nil, fmt.Errorf("registry: creating target dir %q: %w", targetDir, err)
+	}
+
+	res := &fetchResult{}
+	for _, ref := range refs {
```
joycel-github, line 177: let me know if i misread - so this ref is from the curr setup right? what will happen for below?

1. T1: a,b 
2. T2: a with new version 


b will still be there after T2 right? shall we do soemthing like 

stale := old - new 
remove(stale)

internal/skills/geminienterprise/materialize.go
```diff
@@ -0,0 +1,169 @@
...
+//
+// target_dir is a required, validated config field (see config.SkillsConfig), so
+// this does not fall back to any env var or the working directory. It is
+// fail-safe: disabled registries are skipped, and any registry error is logged
+// and swallowed so a skill problem never blocks harness creation. When the same
+// skill id would be written into the same dir more than once (within one
+// registry's selection, or across registries sharing a dir), the FIRST writer
+// wins and later duplicates are skipped with a warning. Substrate/pod
+// materialization is a separate, later path; this wires the local flow only.
+func Materialize(ctx context.Context, sc config.SkillsConfig) Result {
```
joycel-github, line 80: can we have a follow up to add deadline exceeds here? understood the API doens't support long running. maybe we can put a default 120s and let user set

internal/skills/geminienterprise/unzip.go
zbl94: Updated the code to respect the original permission provided from the API. I think this should be the most safe thing to do.

internal/skills/geminienterprise/materialize.go
joycel-github: also do we have concurrent download / refresh? if not, we can add a TODO to consider that to speed up

internal/config/config.go
```diff
@@ -105,6 +110,101 @@ type AntigravityHarnessConfig struct {
...
+	// is written to <TargetDir>/<skill-id>/. Required when Enabled.
+	TargetDir string `yaml:"target_dir,omitempty"`
+}
+
+// SkillsQueryConfig selects skills by semantic search.
+type SkillsQueryConfig struct {
+	// Text is the semantic search string (required for query selection).
+	Text string `yaml:"text"`
+	// TopK bounds the number of matches (<=0 uses the server default).
+	TopK int `yaml:"top_k,omitempty"`
```
joycel-github, line 201: super nit: good to validate per API max

internal/skills/geminienterprise/client.go
zbl94: Your understanding is correct. The failed skill will be logged, but it will not block the harness from starting.  I think to fail is better than silently using a previous skill happen to be in the same path.

internal/skills/geminienterprise/client.go
zbl94: Does T mean turn? If so, the skills will not load in the middle of a conversation, but only at the start of a conversation. There are some discussions here about this decision - https://docs.google.com/document/d/1tPVTTkLPK3r7p_DJTetPl3QOfzbT74IEE0auK4CxywQ/edit?tab=t.0#bookmark=id.itz8yjxzp0ya

internal/skills/geminienterprise/materialize.go
zbl94: Ah, this is already validated in config.go. No worries.

In the validate function, we assert that exactly one of the 3 possible modes should be enabled. All should be enabled explicitly via "all: true"

internal/config/config.go
zbl94: imo I think we can defer to the server side validation logic on a max TopK.

internal/skills/geminienterprise/client.go
zbl94: Why it should happen at the end? We are trying to claim a skill before fetching and writing to file to avoid conflicts (and save API calls if there are)

internal/skills/geminienterprise/materialize.go
zbl94: Added TODOs for both.

nxgczidjiwucj-collab:
Khohjkvxsjgfjjaxgqjqchakca

PR #303, opened 2026-07-14
Honor view_file line range and cap result size

zbl94:
execViewFile read the whole file and ignored the agent's StartLine/EndLine, so viewing a large file returned its entire content as the tool result -- a ~900KB blob for a 3k-line CSV stalled the following turn.

Implement the Antigravity view_file contract (1-indexed inclusive line range with slice-notation windowing) plus defensive caps:
- StartLine/EndLine window (neither=first N lines; start-only=next N forward; end-only=previous N backward; both=precise range, capped to N).
- viewFileMaxLines / viewFileMaxBytes caps so a large file can never blob.

Adds intArg/intArgOK helpers and tests for windowing, byte cap + offset resume, and range honoring.

### DIFF
--- a/internal/harness/antigravityinteractions/antigravityinteractions_tools.go
+++ b/internal/harness/antigravityinteractions/antigravityinteractions_tools.go
@@ -24,6 +24,7 @@ import (
 	"strconv"
 	"strings"
 	"time"
+	"unicode/utf8"
 )
 
 // executeTool runs a tool call the agent yielded and returns the result value to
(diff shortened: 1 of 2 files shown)


wjjclaud (commented):
I don't see where next_content_offset is reported. Is this intentional?

zbl94:
> I don't see where next_content_offset is reported. Is this intentional?

It turns out that this field, `next_content_offset` was just came up with by the coding LLM. It may not be consumed by the interactions API. I asked it to remove. Updated the PR description as well.

internal/harness/antigravityinteractions/antigravityinteractions_tools.go
```diff
@@ -84,7 +95,90 @@ func execViewFile(call capturedToolCall) any {
...
+		// Window is entirely past EOF (or inverted after clamping): nothing to show.
+		return ""
+	}
+
+	return strings.Join(lines[lo-1:hi], "\n")
+}
+
+// applyByteWindow returns the slice of content starting at offset (the agent's
+// ContentOffset), capped to viewFileMaxBytes so a large window can't blob.
+func applyByteWindow(content string, offset int) string {
```
joycel-github, line 200: shall we also ensure it cap at a valid UTF-8 boundary

internal/harness/antigravityinteractions/antigravityinteractions_tools.go
```diff
@@ -84,7 +95,90 @@ func execViewFile(call capturedToolCall) any {
...
+	// a large file can't blob; the returned payload is just the content slice (the
+	// result schema for view_file is defined server-side, so we don't add fields).
+	start, startSet := intArgOK(call.arguments, "StartLine")
+	end, endSet := intArgOK(call.arguments, "EndLine")
+	offset := intArg(call.arguments, "ContentOffset")
+
+	windowed := resolveLineWindow(string(data), start, startSet, end, endSet)
+	content := applyByteWindow(windowed, offset)
+
+	return map[string]any{"content": content}
```
joycel-github: iirc, truncation is silent here? then the model cannot ditinguish a complete read from a truncated one 

we need to return continuation medatadata

internal/harness/antigravityinteractions/antigravityinteractions_tools_test.go
```diff
@@ -0,0 +1,217 @@
...
+		t.Run(tt.name, func(t *testing.T) {
+			got := resolveLineWindow(content, tt.start, tt.startSet, tt.end, tt.endSet)
+			if got != tt.want {
+				t.Errorf("content = %q, want %q", got, tt.want)
+			}
+		})
+	}
+}
+
+func TestResolveLineWindow_StartOnlyForwardWindow(t *testing.T) {
```
joycel-github, line 63: i would also add some test cases for the case mentioned in comments above so 1) continuation cases 2) utf-8 char crossing byte limit

internal/harness/antigravityinteractions/antigravityinteractions_tools.go
zbl94: Done

internal/harness/antigravityinteractions/antigravityinteractions_tools.go
zbl94: It turns out that the server side needs to be changed as well. See cl/948154124. but yes, more fields need to be returned other than content.

internal/harness/antigravityinteractions/antigravityinteractions_tools_test.go
zbl94: done

wjjclaud (approved):

joycel-github (approved):

PR #304, opened 2026-07-14
antigravity: stop exposing harness state_dir in ax.yaml

joycel-github:
remove state_dir from ax.yaml for both antigravity and interaction harness. 

this is to resolve the follow-up from https://github.com/google/ax/pull/292#discussion_r3562628246

### DIFF
--- a/cmd/ax/internal/cliutil/cliutil.go
+++ b/cmd/ax/internal/cliutil/cliutil.go
@@ -71,9 +71,14 @@ func NewControllerFromConfig(ctx context.Context, cfg *Config) (*controller.Cont
 		if address == "" {
 			address = "127.0.0.1:50053"
 		}
-		// Local mode: the harness owns the Python sidecar. Empty StateDir
-		// means the sidecar applies its own default
-		antigravityHarness, err = antigravity.New(ctx, address, cfg.Harnesses.Antigravity.StateDir, true)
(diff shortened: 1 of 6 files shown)


wjjclaud (approved):

cmd/ax/internal/cliutil/cliutil.go
```diff
@@ -97,12 +102,11 @@ func NewControllerFromConfig(ctx context.Context, cfg *Config) (*controller.Cont
 		if agent == "" {
 			agent = antigravityinteractions.DefaultAgent
 		}
-		stateDir := cfg.Harnesses.AntigravityInteractions.StateDir
-		if stateDir == "" {
-			stateDir, err = antigravityinteractions.DefaultStateDir()
-			if err != nil {
-				return nil, fmt.Errorf("antigravity-interactions harness: %w", err)
-			}
+		// AX owns the resume-cursor path (an SDK implementation detail), so
```
zbl94: nit: "an SDK implementation detail" => "a harness implementation detail"? This is not the SDK path.

zbl94 (approved):

cmd/ax/internal/cliutil/cliutil.go
joycel-github: good catch. done

PR #305, opened 2026-07-14
Build separate python image for antigravity harness

wjjclaud:
Split the single combined `ax` image into two images:
- **`ax`** — Used by the ax-server and the Go interactions harness.
- **`ax-antigravity`** —  Used by the antigravity harness built with python SDK.

The resumption time of interactions actor reduce from ~800ms to ~500ms with the removed python dependencies. The resumption time of the antigravity python actor doesn't change much.

#### Tested
Deployed on substrate and run with both antigravity and antigravity-interactions.
Antigravity actor resume: ~1.6s
Interactions actor resume: ~500ms

### DIFF
--- a/cmd/ax/Dockerfile
+++ b/cmd/ax/Dockerfile
@@ -12,14 +12,19 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# Comprehensive AX image: the Go `ax` binary plus the Antigravity Python sidecar
-# (SDK, localharness, agent), on a Debian-based image with shell tooling for
-# skills/code execution. Used by both the ax-server (`ax serve`) and the harness
-# actor (`ax harness`, which forks the Python sidecar).
+# Multi-target AX image build:
+#   --target antigravity : python:3.13-slim + the Antigravity Python sidecar
+#                          (SDK, localharness, agent). Used by the antigravity
+#                          harness actor (`ax harness antigravity`).
+#   --target ax          : debian:stable-slim + the Go `ax` binary. Used by 
+#                          the ax-server (`ax serve`) and the Go interactions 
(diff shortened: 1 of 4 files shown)


cmd/ax/Dockerfile
```diff
@@ -12,14 +12,19 @@
...
+#                          the ax-server (`ax serve`) and the Go interactions 
+#                          harness actor (`ax harness antigravity-interactions`).
 #
-# Build context: repository root. The image targets the cluster's linux/amd64
-# nodes, so it is built with `--platform linux/amd64`. For a standalone build:
-#   docker build --platform linux/amd64 -f cmd/ax/Dockerfile -t <ref> . && docker push <ref>
+# Build context: repository root. Images target the cluster's linux/amd64 nodes,
+# so build with `--platform linux/amd64`. 
+# For a standalone build:
+#   docker build --platform linux/amd64 --target antigravity -f cmd/ax/Dockerfile -t <ref> . && docker push <ref>
```
rakyll, line 26: Do we need this? You can use the same image, with different startup commands.

I'd start with keeping everything in a single image, using actor template to provide custom startup commands. We can publish a secondary image as an optimization -- e.g. a smaller image.

wjjclaud: We already have different startup commands in the actor templates that work e2e. The separate python image is only for optimization purpose that I want to propose and hear your thoughts.
For interactions, we could save 300ms on the interactions path with smaller image size if we strip the python dependencies. I think it's good for us to establish a performance baseline for interactions harness.

rakyll: Makes sense, I forgot about the fact that Python dependencies were huge and it's impacting the snapshots.

rakyll (approved):

PR #311, opened 2026-07-15
antigravity format clean up - harness_server.py 

joycel-github:
harness_server.py  is not formatted. When adding logical change, we only get irrelevant format delta and need to prompt to remove. Format this once. There is no real delta


wjjclaud (approved):
Do you mind adding a comment that which python formatter is used so we can configure our IDE with the same?

PR #316, opened 2026-07-15
skills: add local directory source alongside the registry

zbl94:
Introduce a local skills source parallel to the Gemini Enterprise Skill Registry integration. A local source points at a directory already on disk (a parent dir holding <skill-id>/SKILL.md subfolders, the same layout as examples/skills and a registry's target_dir); nothing is fetched and the directory is used as-is.

- New mode-agnostic result shape (internal/skills: Available/Group/Skill) so both sources feed the same consumers without either depending on the other.
- internal/skills/local.Discover enumerates enabled local paths, treating immediate subfolders with a SKILL.md as skills (subfolder name = id). It is fail-safe like registry materialization: a missing/unreadable/empty path is logged and skipped, never blocking harness creation.
- config.SkillsConfig gains Local []LocalSkillsConfig (enabled + path) with validation; geminienterprise.Materialize now returns skills.Available.
- cliutil combines registry and local groups into one system-instruction pointer for the Antigravity Interactions harness. The Antigravity SDK harness (SKILLS_DIR-based) is left as a TODO.

Only the interactions harness is wired; tests cover discovery, fail-safe degradation, config validation, and the examples/skills fixture.

### DIFF
--- a/cmd/ax/internal/cliutil/cliutil.go
+++ b/cmd/ax/internal/cliutil/cliutil.go
@@ -26,7 +26,9 @@ import (
 	"github.com/google/ax/internal/harness/antigravity"
 	"github.com/google/ax/internal/harness/antigravityinteractions"
 	"github.com/google/ax/internal/harness/substrate"
+	"github.com/google/ax/internal/skills"
 	"github.com/google/ax/internal/skills/geminienterprise"
+	"github.com/google/ax/internal/skills/local"
 )
 
 // Controller is the active controller type for this build.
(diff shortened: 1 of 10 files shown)


rakyll:
We caused a bunch of conflicts for this PR. If it's still relevant, can we fix the merge conflicts?

zbl94:
> We caused a bunch of conflicts for this PR. If it's still relevant, can we fix the merge conflicts?

The conflicts are fixed. Please take a look again.

joycel-github (approved):

Issue #299, opened 2026-07-11
Verify README for v0.2.0 release


rakyll:

joycel-github:
- #309 — fix invalid harness in exec example (base: `main`)
- #310 — refresh Antigravity setup, auth, and CLI usage (stacked on #309)

More to come

Issue #262, opened 2026-07-07
ConversationEvent.harness_config should be bytes


rakyll:
AX has no opinionation whether the config can be represented as a struct type. The config serialization and deserialization is a `Harness` implementation detail.

The config could be:
- A struct or a list serialized as JSON
- A proto payload
- A yaml payload
- A payload in a custom configuration format

PR #319, opened 2026-07-16
docs(readme): correct --last-seq disconnect/resume docs

joycel-github:
## Summary
- The README claims the server replays events after `--last-seq` so a disconnected client can catch up. The controller does not read `ExecRequest.LastSeq` (no server-side reader in `internal/`), so that behavior does not happen.
- Quick Start: reconnect with `--resume` + the conversation ID instead of the non-functional `--last-seq` catch-up example.
- Usage: describe `--last-seq` as reserved and note the server does not currently act on it.

## Context
This regressed when the resume path was rewritten (server-side `last_seq` handling existed via #262, dropped in the #308 controller rewrite). Code gap tracked in #320; this PR only corrects the docs.

Part of #299


rakyll:
Closing this PR due to the significant changes made to the proto.

Issue #268, opened 2026-07-07
Integrate DurableDir in AX


wjjclaud:
Substrate introduces DurableDir as a new mechanism for snapshots. This should help improve the overall performance of actor resumption.

For more context: see the tracking issue from substrate https://github.com/agent-substrate/substrate/issues/119.

joycel-github:
moved to milestone 0.3.0

PR #327, opened 2026-07-16
Interactions: persist interactions harness with durable dir and resume from it

wjjclaud:
The Antigravity interactions harness is built upon a thin Go server, which is fast in terms of start up time. Instead of taking a snapshot of the whole rootfs + memory, we can take snapshot of one durable directory only, and resume from the state under that durable dir. Related issue: #268

#### Changes
- The interactions harness write artifacts to `workspace` by default.
- Mount durable dir at `/workspace`, which contains all the artifacts the interactions harness has created, as well as the internal state `.ax/antigravityinteractions/cursor`
- Add MkdirAlls in setHarnessWorkDir before chdir call.

#### Tested
- Deploy on substrate.
- Test with interactions harness: write files, update files, execute files across turns.

### DIFF
--- a/manifests/ax-deployment.yaml
+++ b/manifests/ax-deployment.yaml
@@ -97,7 +97,9 @@ spec:
         - name: GOOGLE_CLOUD_PROJECT
           value: "${GOOGLE_CLOUD_PROJECT}"
         - name: AX_HARNESS_WORKDIR
-          value: "/workspace"
+          value: "/durable/workspace"
+        - name: AX_DURABLE_DIR
+          value: "/durable"
         # Content of manifests/ax.yaml
         - name: AX_CONFIG_CONTENT
           value: "${AX_CONFIG_CONTENT}"
@@ -106,10 +108,10 @@ spec:
           path: /readyz
           port: 8081
       volumeMounts:
-        - name: workspace
-          mountPath: /workspace
+        - name: durable
+          mountPath: /durable
   volumes:
-    - name: workspace
+    - name: durable
       durableDir: {}
   snapshotsConfig:
     location: gs://${AX_SNAPSHOTS_BUCKET}/axinteractions/
--- a/manifests/ax-deployment.yaml
+++ b/manifests/ax-deployment.yaml
@@ -98,6 +98,8 @@ spec:
           value: "${GOOGLE_CLOUD_PROJECT}"
         - name: AX_HARNESS_WORKDIR
           value: "/workspace"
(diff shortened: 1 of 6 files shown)


manifests/ax-deployment.yaml
```diff
@@ -98,15 +98,25 @@ spec:
           value: "${GOOGLE_CLOUD_PROJECT}"
         - name: AX_HARNESS_WORKDIR
           value: "/workspace"
+        - name: AX_DURABLE_DIR
```
rakyll: What's the different between AX_HARNESS_WORKDIR and AX_DURABLE_DIR? Should we use AX_HARNESS_WORKDIR as the durable dir?

rakyll (approved):

manifests/ax-deployment.yaml
joycel-github: Similar question. iiuc 

1. `AX_DURABLE_DIR` becomes the base for AX assets, including the cursor store.
2. `AX_HARNESS_WORKDIR` is the workspace for local tool execution.

Shall we use the following layout so tool execution doesn’t accidentally affect AX’s internal state?

```yaml
volumeMounts:
  - name: workspace
    mountPath: /durable

env:
  - name: AX_DURABLE_DIR
    value: /durable
  - name: AX_HARNESS_WORKDIR
    value: /durable/workspace
```

zbl94: I think using `/durable/.ax/antigravityinteractions/cursor` to store the cursor makes sense.

wjjclaud: I was thinking about something similar to `/durable` and `/durable/workspace`, so I added the AX_DURABLE_DIR. It would work for the SDK harness. 
But for the interactions harness, on the server side it has system instructions that specifically says the `/workspace` directory. I tried changing the workspace to something like `/durable/workspace`, but it always tries to write to `/workspace`, which won't live through turns.

I will remove AX_DURABLE_DIR  and just use AX_HARNESS_WORKDIR for the scope of this PR since we only add this for interactions.

zbl94: I think the explicit mentioning of `/workspace` in the system instruction should be already fixed. Can you start a sherlog to verify what is the full system instruction received by the model?

wjjclaud: Adopted the `/durable` mount path in the comments above. I will merge this PR, but we still need #332 as a follow up.

Issue #325, opened 2026-07-16
Adopt native env-var credential support from google-antigravity 0.1.7 and remove the sidecar override


joycel-github:
## Background

The antigravity harness sidecar (`python/antigravity/harness_server.py`) currently carries a temporary override, `_vertex_kwargs_from_env()`, that reads the standard GenAI environment variables (`GOOGLE_GENAI_USE_VERTEXAI` / `GOOGLE_GENAI_USE_ENTERPRISE` + `GOOGLE_CLOUD_{PROJECT,LOCATION}`) and threads them into `LocalAgentConfig` explicitly.

This override existed because older `google-antigravity` releases (through 0.1.5, which we pin) did **not** read those env vars natively: a user following the documented recipe (export env vars, construct a bare `LocalAgentConfig`) would fall through to the AI Studio path and hit a misleading `"A Gemini API key is required"` error instead of routing to Vertex.

The upstream SDK has since been fixed to read these env vars natively (`LocalAgentConfig.__init__` hydrates `vertex`; `VertexEndpoint` hydrates `project`/`location` via a `model_validator`). That fix shipped in **`google-antigravity==0.1.7`** (published 2026-07-16), verified present in the published wheel.

The override is now redundant and marked `TODO: remove once AGY reads these env vars natively`.

## Plan

Tracked as two small, sequenced PRs:

- [x] **PR 1 — Bump `google-antigravity` 0.1.5 → 0.1.7.** Just `python/antigravity/requirements.txt`. Independently valuable; with the override still present, behavior is unchanged (explicit kwargs from the override simply win, same result).
- [ ] **PR 2 — Remove the `_vertex_kwargs_from_env` override.** Depends on PR 1. Deletes the override + `VertexKwargs`, simplifies `_build_default_config()` to a bare `LocalAgentConfig(system_instructions=...)`, and updates `_has_credentials()` to read `GOOGLE_CLOUD_{PROJECT,LOCATION}` from env (the SDK now hydrates those onto the `VertexEndpoint`, not the `LocalAgentConfig`, so the startup precheck must consult env rather than `config.{project,location}`). Includes the corresponding `harness_server_test.py` updates.

Ordering matters: PR 1 must merge first. Removing the override while still pinned to 0.1.5 would regress to the original bug.

## Notes on the 0.1.5 → 0.1.7 jump

[...]

PR #328, opened 2026-07-16
antigravity: remove _resolve_localhost /etc/hosts hack

joycel-github:
## Summary
- Remove the `_resolve_localhost()` hack that patched `/etc/hosts` so `localhost` resolved under substrate/gVisor.
- Obsolete as of `google-antigravity` 0.1.7: the SDK's `_connect_websocket` now retries both `localhost` and `127.0.0.1` when dialing localharness (`local_connection.py:993`), so the fallback is handled upstream.

## Testing
- [x] Local: sidecar suite 32/32; booted the real gRPC server and confirmed it serves health without the hack.
- [x] Substrate: verified on Agent Substrate (GKE `lhuan-substrate`, gVisor, no injected `/etc/hosts`). Deployed the antigravity harness image built from this branch with Vertex via Workload Identity (no API key in image). Real end-to-end turn returned a model response (`PR328_OK`); suspend/restore (`RESTORE_OK`) and cold restore after deleting all worker pods (`COLD_OK`) both succeeded. Zero `Failed to connect/initialize WebSocket` or localhost-resolution errors across the run. Confirmed SDK 0.1.7 fallback in the image: `local_connection.py` `for host in ("localhost", "127.0.0.1")`. SDK 0.1.7, substrate `v0.0.0-20260706222328-3cb7433bd8a8`.

Part of #325.

### DIFF
--- a/python/antigravity/harness_server.py
+++ b/python/antigravity/harness_server.py
@@ -450,28 +450,6 @@ def _enhance_config_from_env(config) -> None:
             config.skills_paths.append(skills_dir)
 
 
-def _resolve_localhost():
-    """Ensure `localhost` resolves to 127.0.0.1.
-
-    Substrate actors run under gVisor with no runtime-injected /etc/hosts.
-    The antigravity SDK dials localharness at ws://localhost:<port>/
-    and Python's resolver needs `localhost` in /etc/hosts.
-    """
-    try:
-        try:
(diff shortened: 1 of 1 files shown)


rakyll (approved):
Yay!

zbl94 (approved):

PR #334, opened 2026-07-18
Remove the binary

rakyll:


rakyll:
Can we squash some commits to get rid of this binary being in the history? Binaries really impact the git performance and I'd like to remove this from the history at the cost of losing a few commits by a squash.

joycel-github (approved):

